PR #15, opened 2019-08-15
auth/live: Better JS ServerData "parsing" hack. Fixes #14

### LINKED ISSUE #14: Xbox live login fails
`panic: error obtaining Live token: error decoding flowtoken XML container: EOF`

This error started happening really often, then it went away and now it seems like it happens everytime.

I tried with different accounts and also different IPs in case Microsoft blocked my IP because of suspicious login attempts.
I also tried creating a new account using a new residential IP and then use the same IP to login with gophertunnel API, it's not working, same error but I'm able to login using minecraft

gioiann:


Sandertv:
Looks good, thank you!

PR #37, opened 2020-07-03
feat: expose (*raknet.Conn).Latency() method as (*minecraft.Conn).Latency()

Encritary:


Sandertv:
I don't think using the Latency method is a very good thing because it gives such inconsistent numbers. I'd be happier if this used the `rakConn.recoveryQueue.AvgDelay()` method, but that is not exposed yet.

Sandertv:
I have changed go-raknet to do what I said above. Just to make sure this doesn't come back to bite, could you, instead of returning -1, just panic if the connection is not a *raknet.Conn and change the documentation to reflect the new value returned by raknet.Conn.Latency()?

PR #41, opened 2020-08-06
Add other level_event entries

Aericio:
Only BC break is that EventGuardianCurse was changed to EventParticleSoundGuardianCurse because in source it was `ParticleSoundGuardianGhost = 2006`. I thought it would be better to stay consistent. I can undo though.


Sandertv:
Thanks for the PR!

PR #46, opened 2020-10-06
added timeout

Eren5960:


Sandertv:
I absolutely see the need for this but would rather approach this with a new function where a context can be passed.

PR #52, opened 2020-12-08
1.16.200 support

TwistedAsylumMC:


TwistedAsylumMC:
@Sandertv are you able to review this PR?

minecraft/conn.go
```diff
@@ -637,19 +637,25 @@ func (conn *Conn) handleClientToServerHandshake() error {
...
+				// One of the resource packs has scripts, so we set HasScripts in the packet to true.
+				pk.HasScripts = true
+				behaviourPack.HasScripts = true
+			}
+			pk.BehaviourPacks = append(pk.BehaviourPacks, behaviourPack)
 			continue
 		}
-		pk.TexturePacks = append(pk.TexturePacks, resourcePack)
+		texturePack := protocol.TexturePackInfo{UUID: pack.UUID(), Version: pack.Version(), Size: uint64(pack.Len())}
+		if pack.HasScripts() {
```
Sandertv, line 653: Can remove this as texture packs are never expected to have scripts.

Sandertv (commented):
One small thing to change. Documentation on the new packet and fields are still mostly missing.

minecraft/protocol/item_stack.go
```diff
@@ -152,6 +152,8 @@ type StackResponseSlotInfo struct {
 	Count byte
 	// StackNetworkID is the network ID of the new stack at a specific slot.
 	StackNetworkID int32
+	// CustomName ...
+	CustomName string
```
T14Raptor, line 156: What happened to writing this field?

minecraft/protocol/resource_pack.go
```diff
@@ -1,30 +1,57 @@
...
+	ContentKey string
+	// SubPackName ...
+	SubPackName string
+	// ContentIdentity ...
+	ContentIdentity string
+	// HasScripts specifies if the texture packs has any scripts in it. A client will only download the
+	// behaviour pack if it supports scripts, which, up to 1.11, only includes Windows 10.
+	HasScripts bool
+	// RTXEnabled ...
+	RTXEnabled bool
```
Sandertv, line 52: This bool can use documentation. Resource packs can be RTX resource packs so that ray tracing may be enabled when they're applied. This specifies if the pack is one of those.

minecraft/protocol/item_stack.go
```diff
@@ -152,6 +152,8 @@ type StackResponseSlotInfo struct {
 	Count byte
 	// StackNetworkID is the network ID of the new stack at a specific slot.
 	StackNetworkID int32
+	// CustomName ...
```
Sandertv, line 155: Documenting that this is the custom name of the new item stack and that it has something to do with text filtering here should be sufficient.

Sandertv (changes requested):
Just two documentation things, ready to merge after those.

Sandertv (approved):
Looks good, thank you!

PR #54, opened 2021-01-19
Add a simple multiprotocol accept protocols list.

Jviguy:
This PR adds a simple map with keys of integers and a byte type value that can be anything as its unused.


Sandertv:
Going to close this due to reasons previously discussed in the Discord. If multiple protocols are allowed, we must at least support a way to handle different protocols.

PR #69, opened 2021-04-13
1.16.220 support

JustTalDevelops:
Full support for the 1.16.220 update. Testers and feedback would be appreciated.


minecraft/protocol/item.go
```diff
@@ -14,27 +17,127 @@ type ItemInstance struct {
...
+		x.MetadataValue, x.Count, x.CanBePlacedOn, x.CanBreak = 0, 0, nil, nil
+		return
+	}
+	_, isReader := r.(*Reader)
+
+	r.Uint16(&x.Count)
+	r.Varuint32(&x.MetadataValue)
+
+	// This will be overridden if the IO is a Reader anyways.
+	hasNetID := i.StackNetworkID != 0
```
dktapps, line 36: seems like a better idea to just say := false, or leave it uninitialized ..

JustTalDevelops:
Separated the item instance readers and writers. Anything else I should do?

Sandertv (approved):
Thanks for the PR! All looks good.

Issue #83, opened 2021-06-29
Event packets not being forwarded


racerxdl:
Currently the Event packets are not implemented, and since they're length varying packets the decoder sometimes returns a error/invalid packet because of unconsumed data.

This is mostly harmless, but I found out in long gameplays it tends to create some weird behaviours. Locally I fixed by adding a `ExtraBytes` field to the packet to receive all extra bytes. This way, besides GopherTunnel not interpreting the packets, it can be correctly forwarded to the client.

PR #84, opened 2021-06-29
Fix event packets not being parsed correctly

racerxdl:
See #83 

This allows the event packet to be correctly forwarded even if we don't actually implement it. That also allows someone to parse get events by id like player death (which is sent only to the player that died) or mob death.


Sandertv:
Resolved by #93, going to close this. Thanks for your PR anyway!

PR #85, opened 2021-06-29
Add IdentityPassthrough for "transparent proxy"

racerxdl:
When doing proxy for real servers, it might be good to just passthrough the Xbox Identity. This allows for passing through Xbox data.

The target server must have auth disabled, but if Gophertunnel has auth enabled, the fields will be populated correctly, and the clients will see as if the server is correctly authenticated.

### DIFF
--- a/minecraft/dial.go
+++ b/minecraft/dial.go
@@ -59,6 +59,12 @@ type Dialer struct {
 	// server will send chunks as blobs, which may be saved by the client so that chunks don't have to be
 	// transmitted every time, resulting in less network transmission.
 	EnableClientCache bool
+
+	// KeepXBLIdentityData, if set to true, enables passing XUID and other Identity data to target server
+	// if the authentication token has not been set. This is only know to work with BDS and client might
+	// be kicked in other servers due sending a XUID without any authentication.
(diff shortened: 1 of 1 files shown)


Sandertv:
I take it this is for BDS? Main reason this is a bit problematic is because all other server software will not accept it because it's not technically valid.

Sandertv:
I suppose if we document that clearly that might not be an issue though.

racerxdl:
Yeah @Sandertv, BDS works fine, but if to work correctly with Xbox Gaming (achievements, profile, skins) it needs to have authentication enabled in gopher tunnel, and gopher tunnel and send the XUID from the user that connected to BDS (so BDS can forward for other users)

Also it is tecnically valid, because Identity only says XUID it doesnt have an authentication token.

Sandertv:
I'm happy to merge this if you change the field name slightly, to something like `KeepXBLIdentityData`, to reflect what it does a little more clearly. Also please document that setting this to true and not authenticating will only work on BDS. On other server software it will generally kick your join request.

Sandertv:
Thanks for the PR!

PR #91, opened 2021-07-12
Add support for encrypted resource packs

TwistedAsylumMC:
This pull request does NOT implement encryption, but allows encrypted resource packs to be sent via gophertunnel.


Sandertv (approved):
Looks good to me, thank you!

PR #93, opened 2021-07-29
Handle events in the event packet

### LINKED ISSUE #83: Event packets not being forwarded

JustTalDevelops:
Fixes #83.

### DIFF
--- a/minecraft/protocol/command.go
+++ b/minecraft/protocol/command.go
@@ -14,7 +14,7 @@ type Command struct {
 	// a command.
 	Description string
 	// Flags is a combination of flags not currently known. Leaving the Flags field empty appears to work.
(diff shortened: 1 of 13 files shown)


Sandertv (approved):
Lots of missing documentation, but we can work on that later. Thanks a lot for this, it is very much appreciated!

PR #110, opened 2021-12-05
Update most packet enums

### LINKED ISSUE #109: Update enums as most are outdated
Many of the enums, mainly those in LevelEvent and LevelSoundEvent, are out of date by many major versions. This isn't super major, but can be quite irritating and would be helpful for everyone if all were up to date.

JustTalDevelops:
Closes #109.


Sandertv (approved):
Looks good. This is a pretty breaking change but I think it's easy enough to fix for people.

PR #111, opened 2021-12-27
Implement a way to set custom biome definitions

JustTalDevelops:


minecraft/listener.go
```diff
@@ -43,6 +43,9 @@ type ListenConfig struct {
 	// This field should not be edited during runtime of the Listener to avoid race conditions. Use
 	// Listener.AddResourcePack() to add a resource pack after having called Listener.Listen().
 	ResourcePacks []*resource.Pack
+	// BiomeDefinitions is a list of encoded biome definitions that the server may use. If these are nil,
+	// the default biome definitions will be used.
+	BiomeDefinitions []byte
```
Sandertv, line 48: I'd make this a map[string]interface{} and leave the NBT encoding to gophertunnel

JustTalDevelops: Changed.

minecraft/listener.go
```diff
@@ -43,6 +43,9 @@ type ListenConfig struct {
 	// This field should not be edited during runtime of the Listener to avoid race conditions. Use
 	// Listener.AddResourcePack() to add a resource pack after having called Listener.Listen().
 	ResourcePacks []*resource.Pack
+	// BiomeDefinitions is a list of encoded biome definitions that the server may use. If these are nil,
+	// the default biome definitions will be used.
+	BiomeDefinitions map[string]interface{}
```
Sandertv, line 48: I think it makes more sense to put this in the GameData.

Sandertv: Would also just call it `Biomes` there.

JustTalDevelops: Isn't GameData only meant for data in the start game packet?

Sandertv: Not necessarily, although now that you mention it, it does seem to complicate things a bit if we have it contain data of other packets too... I'd just call it `Biomes` anyway and leave it here for now. That's all right.

JustTalDevelops: Alright, will do.

PR #124, opened 2022-03-11
is not byte array but String type

SenseiTarzan:


Sandertv:
This is intentionally a []byte. In Go json data is always represented as a `[]byte` and it prevents copying of the original data. Why do you reckon this should be a string?

Prim69:
If you wanna use it as a string you can cast it with string(formdata)

JustTalDevelops:
Closing this PR because of the reasons discussed above.

PR #130, opened 2022-05-28
Implement 1.19 support

JustTalDevelops:


minecraft/protocol/packet/request_ability.go
```diff
@@ -0,0 +1,80 @@
...
+		r.Bool(&boolVal)
+		r.Float32(&floatVal)
+		pk.Value = boolVal
+	case 2:
+		r.Bool(&boolVal)
+		r.Float32(&floatVal)
+		pk.Value = floatVal
+	default:
+		r.InvalidValue(valType, "ability value type", "must be bool or float32")
+	}
```
DaPigGuy, line 79:
```suggestion
	r.Bool(&boolVal)
	r.Float32(&floatVal)
	switch valType {
	case 1:
		pk.Value = boolVal
	case 2:
		pk.Value = floatVal
	default:
		r.InvalidValue(valType, "ability value type", "must be bool or float32")
	}
```

minecraft/protocol/packet/player_action.go
```diff
@@ -16,6 +16,9 @@ type PlayerAction struct {
 	// BlockPosition is the position of the target block, if the action with the ActionType set concerned a
 	// block. If that is not the case, the block position will be zero.
 	BlockPosition protocol.BlockPos
+	// ResultPosition is the resulting position of the target block. It is unclear what the difference between this
+	// field and BlockPosition is.
+	ResultPosition protocol.BlockPos
```
Sandertv, line 21: I think I can explain what this is for: When a UseItemOn action is sent, the BlockPosition is the block clicked, but when a block is placed, this ResultPosition is the position at which the block will be placed.

Sandertv: Would add that to the documentation.

Sandertv (commented):
One comment, should be good otherwise!

minecraft/protocol/packet/player_action.go
JustTalDevelops: Added.

PR #137, opened 2022-07-11
Allow arrays to be decoded into slices for easier management in structs

JustTalDevelops:


Sandertv:
I'm a bit hesitant on this because it will make decode and encode non-symmetric.

Sandertv:
Going to close this because this would make a decode -> re-encode result in different data than before (different types of tags) which just isn't helpful.

PR #140, opened 2022-07-13
add IsEditorMode to client data (1.19.10)

UnknownOre:


minecraft/protocol/login/data.go
```diff
@@ -105,6 +105,8 @@ type ClientData struct {
 	// GUIScale is the GUI scale of the player. It is by default 0, and is otherwise -1 or -2 for a smaller
 	// GUI scale than usual.
 	GUIScale int `json:"GuiScale"`
+	// IsEditorMode is a value to dictate if the player is in editor mode
```
Sandertv, line 108: Make sure to add the dot here.

minecraft/protocol/login/data.go
```diff
@@ -105,6 +105,8 @@ type ClientData struct {
 	// GUIScale is the GUI scale of the player. It is by default 0, and is otherwise -1 or -2 for a smaller
 	// GUI scale than usual.
 	GUIScale int `json:"GuiScale"`
+	// IsEditorMode is a value to dictate if the player is in editor mode.
+	IsEditorMode bool `json:"IsEditorMode"`
```
Sandertv, line 109: You don't need to add the `json:` tag here if the name is the same as the field name.

Sandertv (approved):
Thanks, looks good!

PR #148, opened 2022-08-26
Fix PlayerDiedEventData

egor-erm:
I use gophertunnel as a proxy server. When a player dies, an error message arrives (*packet.Event: 2 unread bytes left: 0x0a00). I found a mismatch of PlayerDiedEventData with the original protocol. After the correction, no errors occur.

https://github.com/CloudburstMC/Protocol/blob/f92336f2433875a1116c103dc0100b92cd260e2b/bedrock/bedrock-common/src/main/java/com/nukkitx/protocol/bedrock/data/event/PlayerDiedEventData.java

### DIFF
--- a/minecraft/protocol/events.go
+++ b/minecraft/protocol/events.go
@@ -119,15 +119,7 @@ type PlayerDiedEventData struct {
 }
 
 // Marshal ...
-func (p *PlayerDiedEventData) Marshal(w *Writer) {
-	w.Varint32(&p.AttackerEntityID)
-	w.Varint32(&p.AttackerVariant)
-	w.Varint32(&p.EntityDamageCause)
-	w.Bool(&p.InRaid)
-}
-
-// Unmarshal ...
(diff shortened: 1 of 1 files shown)


minecraft/protocol/events.go
```diff
@@ -110,14 +110,28 @@ func (c *CauldronUsedEventData) Marshal(r IO) {
...
+	AttackerVariant int32
 	// EntityDamageCause ...
 	EntityDamageCause int32
+	// InRaid ...
+	InRaid bool
 }
 
 // Marshal ...
-func (p *PlayerDiedEventData) Marshal(r IO) {
+func (p *PlayerDiedEventData) Marshal(w *Writer) {
```
Sandertv, line 122: I don't think this compiles. Implementations of the `protocol.EventData` interface must (only) implement a `Marshal(r IO)` interface.

egor-erm: Oh, I didn't notice that

Sandertv (approved):
Looks good, thank you for the PR!

PR #150, opened 2022-09-02
fix waxing event panicing

olebeck:
Fixes a panic because of the EventData for waxing not being in the switch,
also makes the names of the events for waxing consistent with the rest.

### DIFF
--- a/minecraft/protocol/packet/event.go
+++ b/minecraft/protocol/packet/event.go
@@ -122,8 +122,6 @@ func (pk *Event) Unmarshal(r *protocol.Reader) {
 		pk.EventData = &protocol.WaxedOrUnwaxedCopperEventData{}
 	case EventTypeSneakCloseToSculkSensor:
 		pk.EventData = &protocol.SneakCloseToSculkSensorEventData{}
-	default:
-		pk.EventData = &protocol.UnknownEventData{}
 	}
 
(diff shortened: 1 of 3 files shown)


olebeck:
maybe related bug, the particles when waxing or scraping dont show up just using proxy, even though the sound plays.

minecraft/protocol/packet/event.go
```diff
@@ -118,6 +118,12 @@ func (pk *Event) Unmarshal(r *protocol.Reader) {
 		pk.EventData = &protocol.MovementCorrectedEventData{}
 	case EventTypeExtractHoney:
 		pk.EventData = &protocol.ExtractHoneyEventData{}
+	case EventTypePlayerWaxedOrUnwaxedCopper:
+		pk.EventData = &protocol.WaxedOrUnwaxedCopperEventData{}
+	case EventTypeSneakCloseToSculkSensor:
+		pk.EventData = &protocol.SneakCloseToSculkSensorEventData{}
+	default:
+		pk.EventData = &protocol.UnknownEventData{}
```
Sandertv, line 126: I would rather still panic here (it should be picked up by the `defer recover()` later on) so that we see errors when they occur.

Sandertv (approved):
All looks good now. Thank you for the PR!

PR #152, opened 2022-09-13
protocol/player.go: Fix documentation

cqdetdev:
This PR address important fixes to the documentation in `protocol/player.go`, without which would completely break gophertunnel as a whole.


JustTalDevelops:
Thanks for the PR!

PR #153, opened 2022-09-24
added minecraft.DefaultProtocol()

Blackjack200:
useful for multi-protocol API

### DIFF
--- a/minecraft/protocol.go
+++ b/minecraft/protocol.go
@@ -44,4 +44,4 @@ func (p proto) ConvertFromLatest(pk packet.Packet, _ *Conn) []packet.Packet {
 
 // DefaultProtocol is the Protocol implementation using as default, In default it is current protocol, version and packet
 // pool and does not convert any packets, as they are already of the right type.
-var DefaultProtocol Protocol = proto{}
+var DefaultProtocol = proto{}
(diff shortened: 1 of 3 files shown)


Superomarking:
fr !!

Sandertv (commented):
Thanks for the PR! This is a good idea. I've added one thing I'd recommend changing.

minecraft/protocol.go
```diff
@@ -41,3 +41,9 @@ func (p proto) ConvertToLatest(pk packet.Packet, _ *Conn) []packet.Packet { retu
 func (p proto) ConvertFromLatest(pk packet.Packet, _ *Conn) []packet.Packet {
 	return []packet.Packet{pk}
 }
+
+// DefaultProtocol return the default Protocol implementation, It returns the current protocol, version and packet pool and does not
+// convert any packets, as they are already of the right type.
+func DefaultProtocol() Protocol {
```
Sandertv, line 47: I would change this to `var DefaultProtocol proto` instead.

minecraft/protocol.go
```diff
@@ -41,3 +41,7 @@ func (p proto) ConvertToLatest(pk packet.Packet, _ *Conn) []packet.Packet { retu
 func (p proto) ConvertFromLatest(pk packet.Packet, _ *Conn) []packet.Packet {
 	return []packet.Packet{pk}
 }
+
+// DefaultProtocol is the Protocol implementation using as default, In default it is current protocol, version and packet
+// pool and does not convert any packets, as they are already of the right type.
+var DefaultProtocol Protocol = proto{}
```
Sandertv, line 47: DefaultProtocol can be assigned to now, which I don't think is what we'd want. I'd just make this `var DefaultProtocol proto`, that way it can be used but not assigned to.

Blackjack200: ok

Sandertv (approved):
Looks good, thank you!

PR #155, opened 2022-10-02
Add a function to stop packs from being downloaded

TwistedAsylumMC:


minecraft/dial.go
```diff
@@ -54,6 +54,11 @@ type Dialer struct {
 	// from which the packet originated, and the destination address.
 	PacketFunc func(header packet.Header, payload []byte, src, dst net.Addr)
 
+	// DownloadResourcePack is called individually for every texture and behaviour pack sent by the connection when
+	// using Dialer.Dial(), and can be used to stop the pack from being downloaded. The function is called with the UUID
+	// and version of the resource pack, and the boolean returned determines if the pack will be downloaded or not.
+	DownloadResourcePack func(packUUID uuid.UUID, version string) bool
```
Sandertv, line 60: Would rename `packUUID` as `id` but otherwise this is ready to go I think.

minecraft/conn.go
```diff
@@ -835,6 +841,9 @@ func (conn *Conn) handleResourcePacksInfo(pk *packet.ResourcePacksInfo) error {
 			conn.packQueue.packAmount--
 			continue
 		}
+		if conn.downloadResourcePack != nil && !conn.downloadResourcePack(uuid.MustParse(pack.UUID), pack.Version) {
+			continue
```
JustTalDevelops, line 845: Shouldn't you decrement packAmount here?

PR #158, opened 2022-11-16
Minor correction

DJStompZone:
Added missing text packet type `TextTypeTranslation` to doc comments for `NeedsTranslation` and `Parameters`

### DIFF
--- a/minecraft/protocol/packet/text.go
+++ b/minecraft/protocol/packet/text.go
@@ -27,7 +27,7 @@ type Text struct {
 	TextType byte
 	// NeedsTranslation specifies if any of the messages need to be translated. It seems that where % is found
 	// in translatable text types, these are translated regardless of this bool. Translatable text types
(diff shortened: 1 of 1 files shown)


JustTalDevelops:
That list isn't meant to be a full list of text types, but rather just a sample of the text types, as the 'include' should indicate.

DJStompZone:
> For future reference - that list isn't meant to be a full list of text types, but rather just a sample of the text types, as the 'include' should indicate.

I guess I was thrown off by the [current wording](https://github.com/Sandertv/gophertunnel/blob/6b6965ed44ba0af068ab4096129a1e955996f526/minecraft/protocol/packet/text.go#L38), 

> These parameters are **only** written if the type of the packet is TextTypeTip, TextTypePopup or TextTypeJukeboxPopup.

JustTalDevelops:
Ah, those are in two different sections. It might make sense to update the other section to be more specific.

PR #165, opened 2023-01-02
reverse defaultclientdata and defaultidentitydata

Prim69:
Client data relies on identity data, but the identity data is being set after

so if the user passes a blank DisplayName in then identity data would set it to "Steve", but this would happen after setting the client data, leading to ThirdPartyName and DisplayName mismatching


JustTalDevelops (approved):
Thanks for the PR!

PR #170, opened 2023-01-24
Add total packs to DownloadResourcePack func

Prim69:
Some servers have a crazy amount of packs like Hive at 22, and these take a while to download. I wanna show my users the progress while downloading: `[Resource Packs] - Downloading (1/22)`


minecraft/dial.go
```diff
@@ -56,8 +56,9 @@ type Dialer struct {
 
 	// DownloadResourcePack is called individually for every texture and behaviour pack sent by the connection when
 	// using Dialer.Dial(), and can be used to stop the pack from being downloaded. The function is called with the UUID
-	// and version of the resource pack, and the boolean returned determines if the pack will be downloaded or not.
-	DownloadResourcePack func(id uuid.UUID, version string) bool
+	// and version of the resource pack, and the total amount of packs. The boolean returned determines if the pack will
+	// be downloaded or not.
+	DownloadResourcePack func(id uuid.UUID, version string, totalPacks int) bool
```
JustTalDevelops, line 61: Probably a good idea to also include the current pack index too?

Prim69: That can be tracked by the user but i could if needed

Sandertv: Feels like pretty much any time you'd want the total packs, you'd also want the current index. I don't think adding the current index is a bad idea.

Prim69: done

Sandertv (approved):
I think this is okay like this. Thank you for the PR!

PR #173, opened 2023-02-12
Fix incorrect documentation for client_item_start_cooldown

alvin0319:
client_start_item_cooldown is sent by server to tell the client to start cooldown of given item.

Here is the example case I've implemented in PocketMine(Edited Ender Pearl's cooldown to 15 seconds):

https://user-images.githubusercontent.com/32565818/218313575-11f26518-834e-49c7-8635-3ed7bf0d8c20.mp4


Sandertv (approved):
Thank you! Looks good.

PR #183, opened 2023-04-18
Ability to implement your own reader and writer

Flonja:
**Note:** `ItemInstance` and `Item` both have the `NewReader`/`Writer` and I'd like to hear if there would be a (better) way of handling that

### DIFF
--- a/minecraft/conn.go
+++ b/minecraft/conn.go
@@ -319,7 +319,7 @@ func (conn *Conn) WritePacket(pk packet.Packet) error {
 	l := buf.Len()
 
 	for _, converted := range conn.proto.ConvertFromLatest(pk, conn) {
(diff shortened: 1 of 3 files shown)


Sandertv:
I think this idea makes a lot of sense. I'm not sure how to solve that problem of ItemInstance and Item using NewReader and NewWriter though. Do you reckon that's a big issue?

Flonja:
> I think this idea makes a lot of sense. I'm not sure how to solve that problem of ItemInstance and Item using NewReader and NewWriter though. Do you reckon that's a big issue?

I don't think so, since a developer could reference their own IO if needed. But I'm still talking about the issue anyway.

Sandertv (approved):
Looks good, thank you!

PR #195, opened 2023-08-22
use the mojang new root key

UnknownOre:


TobiasGrether:
That key might be the new one, but that will not fix the issue most likely. It seems they are using an entirely new key to sign old game versions, at least from what we can see right now

JustTalDevelops:
> That key might be the new one, but that will not fix the issue most likely. It seems they are using an entirely new key to sign old game versions, at least from what we can see right now

This issue is not referencing old clients - Gophertunnel just never updated to the new public key, so this should actually work fine. Thanks for the PR!

JustTalDevelops (approved):

PR #197, opened 2023-09-06
pass the correct reader limit value to the reader

UnknownOre:


Sandertv (approved):
Good catch, thank you.

PR #198, opened 2023-09-17
Correct the invalid packets setting.

UnknownOre:
This typo caused disconnections when receiving invalid packets, whether you had it enabled or not.


TwistedAsylumMC:
Thanks for fixing this!

PR #224, opened 2024-04-29
fix packet ids

GameParrot:


TwistedAsylumMC:
Forgot to fix this, thanks for the PR!

PR #229, opened 2024-06-16
minecraft/conn.go: Introduce `ReadBytes`

cooldogedev:
This PR introduces `ReadBytes`, a method for efficiently reading packet payloads in `minecraft/conn.go`. This improvement is particularly useful for scenarios like proxies that forward data directly, where it eliminates unnecessary packet decoding overhead.


TwistedAsylumMC:
Thanks for the PR!

PR #231, opened 2024-06-18
Update pong data

HashimTheArab:
This pr fixes the incorrect server subname index, and adds a 0 at the end of the pong data to match with BDS

https://github.com/PrismarineJS/bedrock-protocol/blob/84c5231b92df9f5f1a09b29a05e7abfed62f1c2b/src/server/advertisement.js#L34

pong data from a bds server
MCPE;Dedicated Server;685;1.21.0;0;10;9799519352691166920;Bedrock level;Survival;1;19133;19134;0;


TwistedAsylumMC:
Do you know what the last 0 is for? If not then we can just merge this for now and look into it in the future

HashimTheArab:
It's always padded on idk what for

Adrian8115:
zero is for nintendo switch limited

PR #232, opened 2024-06-22
Authentication error message improvements

HashimTheArab:
- Fix expiresIn key
- Add xbox error codes gotten from https://github.com/PrismarineJS/prismarine-auth/blob/e70f42efddf01d4edeb3a3bb53c19cb76fd66ad0/src/common/Constants.js#L38 and https://wiki.vg/Microsoft_Authentication_Scheme
- Add error description for poll errors

Some responses from the error_description addition:
-  invalid_grant: The user could not be authenticated or user interaction is required. The user must sign in again and if needed grant the client application access to the requested scope.
- expired_token: The provided value for the input parameter 'device_code' is not valid. This device code has expired.

### DIFF
--- a/minecraft/auth/xbox.go
+++ b/minecraft/auth/xbox.go
@@ -95,8 +95,8 @@ func obtainXBLToken(ctx context.Context, c *http.Client, key *ecdsa.PrivateKey, 
 	}()
 	if resp.StatusCode != 200 {
 		// Xbox returns a custom error code in the x-err header.
-		if xboxErr := parseXboxErrorCode(resp.Header.Get("x-err")); xboxErr != "" {
-			return nil, fmt.Errorf("POST %v: %v", "https://sisu.xboxlive.com/authorize", xboxErr)
+		if errorCode := resp.Header.Get("x-err"); errorCode != "" {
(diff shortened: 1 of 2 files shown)


minecraft/auth/xbox.go
```diff
@@ -191,3 +196,28 @@ func sign(request *http.Request, body []byte, key *ecdsa.PrivateKey) {
...
+	case "2148916235":
+		return "Your account resides in a region that Xbox has not authorized use from. Xbox has blocked your attempt at logging in."
+	case "2148916236":
+		return "Your account requires proof of age. Please login to https://login.live.com/login.srf and provide proof of age."
+	case "2148916237":
+		return "Your account has reached its limit for playtime. Your account has been blocked from logging in."
+	case "2148916238":
+		return "The account date of birth is under 18 years and cannot proceed unless the account is added to a family by an adult."
+	default:
+		return ""
```
TwistedAsylumMC, line 220: I'd make this return something like "unknown error code xyz" since code is never going to be empty, you don't want to return empty errors

HashimTheArab:
Fixed

PR #233, opened 2024-07-01
minecraft/listener.go: add methods for safely managing resource packs

cooldogedev:


Sandertv (commented):
Not sure I'm a huge fan of this idea, because adding/removing resource packs doesn't work for clients already connected.

Edit: As discussed on Discord, with clear documentation this is probably fine.

minecraft/listener.go
```diff
@@ -172,6 +175,26 @@ func (listener *Listener) Disconnect(conn *Conn, message string) error {
...
+}
+
+// RemoveResourcePack removes a resource pack from the listener's configuration by its UUID.
+func (listener *Listener) RemoveResourcePack(uuid string) {
+	listener.mu.Lock()
+	defer listener.mu.Unlock()
+	for i, resourcePack := range listener.cfg.ResourcePacks {
+		if resourcePack.UUID() == uuid {
+			listener.cfg.ResourcePacks[i] = nil
+			listener.cfg.ResourcePacks = append(listener.cfg.ResourcePacks[:i], listener.cfg.ResourcePacks[i+1:]...)
```
Sandertv, line 192: These actions on `listener.cfg.ResourcePacks` are not (thread-)safe. When a `*Listener` is created, it should probably create a copy of the Config's slice to prevent it from keeping the same underlying array.

minecraft/listener.go
```diff
@@ -172,6 +177,28 @@ func (listener *Listener) Disconnect(conn *Conn, message string) error {
...
+func (listener *Listener) RemoveResourcePack(uuid string) {
+	listener.packsMu.Lock()
+	defer listener.packsMu.Unlock()
+	for i, resourcePack := range listener.packs {
+		if resourcePack.UUID() == uuid {
+			listener.packs[i] = nil
+			listener.packs = append(listener.packs[:i], listener.packs[i+1:]...)
+			break
+		}
+	}
```
Sandertv, line 199: Consider using the `slices.DeleteFunc` function for this.

minecraft/listener.go
```diff
@@ -128,6 +132,7 @@ func (cfg ListenConfig) Listen(network string, address string) (*Listener, error
 	listener := &Listener{
 		cfg:      cfg,
 		listener: netListener,
+		packs:    append([]*resource.Pack{}, cfg.ResourcePacks...),
```
Sandertv, line 135: Can use `slices.Clone()` here.

minecraft/listener.go
```diff
@@ -227,14 +254,17 @@ func (listener *Listener) listen() {
 // createConn creates a connection for the net.Conn passed and adds it to the listener, so that it may be
 // accepted once its login sequence is complete.
 func (listener *Listener) createConn(netConn net.Conn) {
+	listener.packsMu.RLock()
+	defer listener.packsMu.RUnlock()
```
Sandertv, line 258: This might not be safe here because `listener.packs` could be changed and read simultaneously (as it's a reference type). Might want to RLock, then `packs := slices.Clone(listener.packs)`, then unlock and pass `packs` to `conn.resourcePacks`.

Sandertv (commented):
Sorry for the late review, got some more comments.

minecraft/listener.go
```diff
@@ -172,6 +178,24 @@ func (listener *Listener) Disconnect(conn *Conn, message string) error {
 	return conn.Close()
 }
 
+// AddResourcePack adds a new resource pack to the listener's resource packs.
+// Note: This methods will not update resource packs for active connections.
```
TwistedAsylumMC, line 182: I'd just make this say "This method" rather than "This methods", same for RemoveResourcePack

PR #241, opened 2024-07-23
Update book_edit.go

smell-of-curry:
- Properly match enum type


TwistedAsylumMC:
Closing this since it's an unnecessary breakage, gophertunnel does not aim to match names 1:1 with vanilla

PR #242, opened 2024-07-23
Update boss_event.go

smell-of-curry:
- Properly map enum names


TwistedAsylumMC:
Closing this since it's an unnecessary breakage, gophertunnel does not aim to match names 1:1 with vanilla

PR #243, opened 2024-07-23
Update completed_using_item.go

smell-of-curry:
- Mapped Enum Names properly and set using equals


TwistedAsylumMC:
Closing this since iota is what gophertunnel uses everywhere and UseItemUnknown/UseItemCount are not needed

PR #244, opened 2024-07-23
fix entity properties not being unmarshaled correctly

olebeck:
looked with a regex, there shouldnt be any other Marshal functions that use by value.
` [^\*]+\) Marshal\(r IO\)` used this regex to check


TwistedAsylumMC:
Thanks for the contribution!

PR #245, opened 2024-07-24
use NetworkBlockPos in BlockChangeEntry

olebeck:
https://mojang.github.io/bedrock-protocol-docs/html/UpdateSubChunkBlocksPacket.html


TwistedAsylumMC:
Good catch!

PR #248, opened 2024-07-29
Update pool.go To Fix Packet 75

smell-of-curry:
When going from the end to the overworld you are prompted with the credits screen, this is sent from the server. However when a player skips the credits the player sends back the credits packet, which is currently not handled.

image


TwistedAsylumMC:
Thanks!

PR #252, opened 2024-08-04
Fix in-proper depreciation message in `client_cheat_ability.go`

smell-of-curry:
Properly Standerdise Deprecated noting to match GOLang standard.


smell-of-curry:
Was merged to https://github.com/Sandertv/gophertunnel/pull/254

PR #253, opened 2024-08-04
Mark `FilterText` packet as deprecated in `filter_text.go`

smell-of-curry:
Properly Standerdise Deprecated noting to match GOLang standard.

I don't know when this was deprecated tho.


smell-of-curry:
Was merged to https://github.com/Sandertv/gophertunnel/pull/254

PR #258, opened 2024-08-11
Implement publishing and joining worlds

### LINKED ISSUE #219: Ability to join xbox friend worlds
Just like how gophertunnel allows you to join realms, friend worlds should be added too. As far as I know, this hasn't been done before, so this will make it the first proxy with this ability.

lactyy:
Adds a support for publishing and joining worlds displayed in the friend list. Closes #219.

WIP. I might separate these packages into a different module to reduce usage of huge dependencies added for WebRTC.

Task Lists:
- [x] Implement listening on NetherNet
- [x] Implement publishing worlds
- [x] Implement dialing on NetherNet
- [x] Implement looking up worlds
- [x] Properly write a documentation on types, methods, fields, etc. (Help needed...)
- [x] Implement LAN discovery


lactyy:
The changes in pull request and other packages, which have been split into separate modules, has been fully documented.

minecraft/franchise/signaling/conn.go
```diff
@@ -0,0 +1,200 @@
...
+		c.notifiersMu.Lock()
+		delete(c.notifiers, i)
+		c.notifiersMu.Unlock()
+	}
+}
+
+// Credentials blocks until [nethernet.Credentials] are received from the server or the [context.Context]
+// is done. It returns a [nethernet.Credentials] or an error if the Conn is closed or the [context.Context]
+// is canceled or exceeded a deadline.
+func (c *Conn) Credentials(ctx context.Context) (*nethernet.Credentials, error) {
```
HashimTheArab, line 77: the comment says this method is blocking but the select has a default case

minecraft/room/status.go
```diff
@@ -0,0 +1,108 @@
...
+	CrossPlayDisabled       bool         `json:"CrossPlayDisabled"`
+	TitleID                 int64        `json:"TitleId"`
+	SupportedConnections    []Connection `json:"SupportedConnections"`
+}
+
+type Connection struct {
+	ConnectionType uint32 `json:"ConnectionType"`
+	HostIPAddress  string `json:"HostIpAddress"`
+	HostPort       uint16 `json:"HostPort"`
+	NetherNetID    uint64 `json:"NetherNetId"`
```
HashimTheArab, line 35: dont forget to change this to a string

minecraft/room/status.go
```diff
@@ -0,0 +1,108 @@
...
+		Protocol:                protocol.CurrentProtocol,
+		BroadcastSetting:        BroadcastSettingFriendsOfFriends,
+		LanGame:                 true,
+		OnlineCrossPlatformGame: true,
+		CrossPlayDisabled:       false,
+		TitleID:                 0,
+	}
+}
+
+func NetherNetID(status Status) (uint64, bool) {
```
HashimTheArab, line 99: would change this to func (status Status) NetherNetID()

minecraft/room/status.go
HashimTheArab: actually its sent as either a string UUID or a uint64, not a string-based uint64 so this needs a custom type which supports `string | uint64` decoding

minecraft/auth/xal/token_source.go
```diff
@@ -0,0 +1,45 @@
...
+	t  *oauth2.Token
+	x  *auth.XBLToken
+	mu sync.Mutex
+}
+
+func (r *refreshTokenSource) Token() (_ xsapi.Token, err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.t == nil || !r.t.Valid() || r.x == nil {
```
HashimTheArab, line 34: this should check x.Valid() but the whole function can be improved like this
```go
func (r *refreshTokenSource) Token() (_ xsapi.Token, err error) {
	r.mu.Lock()
	defer r.mu.Unlock()

	// Return the cached XBL token if it is valid.
	if r.x != nil && r.x.Valid() {
		return r.x, nil
	}

	// Request a new underlying token if it is not valid.
	if r.t == nil || !r.t.Valid() {
		r.t, err = r.underlying.Token()
		if err != nil {
			return nil, fmt.Errorf("request underlying token: %w", err)
		}
	}

	// Request a new XBL token using the underlying token.
	r.x, err = auth.RequestXBLToken(r.ctx, r.t, r.relyingParty)
	if err != nil {
		return nil, fmt.Errorf("request xbox live token: %w", err)
	}

	return r.x, nil
}

```

PR #260, opened 2024-08-11
Fix in-proper `compression.go` docs that don't align with actual implementation

smell-of-curry:


minecraft/protocol/packet/compression.go
```diff
@@ -24,10 +24,10 @@ var (
 	// NopCompression is an empty implementation that does not compress data.
 	NopCompression nopCompression
 	// FlateCompression is the implementation of the Flate compression
-	// algorithm. This was used by default until v1.19.30.
+	// algorithm. This is used by default.
 	FlateCompression flateCompression
 	// SnappyCompression is the implementation of the Snappy compression
-	// algorithm. This is used by default.
+	// algorithm. Currently crashes devices without `avx2`
```
TwistedAsylumMC, line 30: I'd make this say "Snappy currently crashes on devices that do not support avx2." just to be more formal

PR #267, opened 2024-08-20
Improve closed connection error messages

HashimTheArab:
This PR improves error messages when handling packets.
These messages were written to the debugger before instead of being handled as an error

Before:
DEBU[0005] dial attempt 1: dial minecraft 192.168.140.68:64906->162.248.102.67:19132: use of closed connection

After:
DEBU[0005] dial attempt 1 failed: dial minecraft 12.168.140.68:64906->162.248.102.67:19132: handle *packet.PlayStatus: server outdated

The removed lines `_ = conn.Close()` should not make a difference they are already handled immediately in the following places:
https://github.com/Sandertv/gophertunnel/blob/aed4aa48e09ecc81cf35f4e2625d41ff2efd1e52/minecraft/dial.go#L282
https://github.com/Sandertv/gophertunnel/blob/aed4aa48e09ecc81cf35f4e2625d41ff2efd1e52/minecraft/listener.go#L298
The reason for removal is so the select can catch the error before the closed conn

### DIFF
--- a/minecraft/dial.go
+++ b/minecraft/dial.go
@@ -218,7 +218,7 @@ func (d Dialer) DialContext(ctx context.Context, network, address string) (conn 
 		conn.identityData = identityData
 	}
 
-	l, c, e := make(chan struct{}), make(chan struct{}), make(chan error, 1)
+	l, c, e := make(chan struct{}), make(chan struct{}),  make(chan error, 1)
 	go listenConn(conn, d.ErrorLog, l, c, e)
 
 	conn.expect(packet.IDNetworkSettings, packet.IDPlayStatus)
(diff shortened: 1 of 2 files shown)


HashimTheArab:
Im thinking we should only do this for conn.receive and not Decode so we dont end up with errors like this

DEBU[0009] dial attempt 1 failed: dial minecraft 192.168.140.68:64680->139.99.39.244:19132: read batch: read raknet 192.168.140.68:64680->139.99.39.244:19132: use of closed network connection 

but thats not consistent so not sure what to do

smell-of-curry:
nice!

minecraft/dial.go
```diff
@@ -290,12 +294,14 @@ func listenConn(conn *Conn, logger *log.Logger, l, c chan struct{}) {
 			if !errors.Is(err, net.ErrClosed) {
 				logger.Printf("dialer conn: %v\n", err)
 			}
+			e <- err
```
Sandertv, line 297: If this (or the next `e <- err`) is reached after the connection is created (DialContext is done running), it will block the goroutine forever because there's nothing to take the error out of the channel.

Sandertv (changes requested):

minecraft/dial.go
HashimTheArab: i think i (cursor + claude sonnet) fixed it in the latest commit

PR #271, opened 2024-11-16
minecraft/text: Add Resin color

inotflying:


TwistedAsylumMC:
Did not notice they added this, thanks!

PR #277, opened 2024-12-26
Use go-jose/go-jose/v4/ for JWT

Dasciam:


Sandertv (approved):
Looks good, thank you!

PR #278, opened 2024-12-29
Add BitSet.Len()

AkmalFairuz:
BitSet.Len() is needed for multi-protocol support purposes. 

Also in C++, there's `size()` method: https://en.cppreference.com/w/cpp/utility/bitset/size


TwistedAsylumMC:
Thanks!

PR #292, opened 2025-03-17
set default values for defaultinputmode and currentinputmode in clientdata

HashimTheArab:
gophertunnel currently doesnt set a default input mode for client data
this gets detected by anticheats because 0 is an invalid input mode


TwistedAsylumMC:
Thanks!

PR #297, opened 2025-03-25
Update pool.go to fix `unexpected packet (ID=321)`

smell-of-curry:
Fix error: `time=2025-03-25T18:21:09.571-04:00 level=ERROR msg="read packet: unexpected packet (ID=321)" src=listener raddr=127.0.0.1:53589`


TwistedAsylumMC:
Thanks!

PR #308, opened 2025-04-14
Game Version Handling

CJMustard1452:
Implementation of `GameVersion string` to minecraft.ListenConfig

GameVersion is the version string that the server presents to clients during login. It should reflect the version of the game that the server is compatible with in terms of resource packs, world generation, and client rendering. If left empty, the default GameVersion for the current Protocol will be used.

(looks like `go fmt ./...` changed a few styles, I can remove it if needed :) )


CJMustard1452:
This PR will become un-drafted once tested.

CJMustard1452:
this pr is not needed lmao

PR #319, opened 2025-05-24
listener: implemented support for per-player resource pack determination

cooldogedev:


GameParrot:
Should include client data and remote address too

cooldogedev:
> Should include client data and remote address too

How would the remote address be useful in this context?

GameParrot:
> > Should include client data and remote address too
> 
> How would the remote address be useful in this context?

Would be nice to have all information about the player instead of just identity data

cooldogedev:
> Would be nice to have all information about the player instead of just identity data

Client data would be nice to have indeed, it's just that I'm not sure about how the remote address would be any useful in this context

> Or just pass the conn

I didn't do that so it aligns with the current convention e.g. https://github.com/Sandertv/gophertunnel/blob/master/minecraft/listener.go#L80

GameParrot:
> > Would be nice to have all information about the player instead of just identity data
> 
> Client data would be nice to have indeed, it's just that I'm not sure about how the remote address would be any useful in this context
> 
> > Or just pass the conn
> 
> I didn't do that so it aligns with the current convention e.g. https://github.com/Sandertv/gophertunnel/blob/master/minecraft/listener.go#L80

Ip could be useful for sending different resource pack based on things like region

PR #324, opened 2025-06-18
protocol/login: Fixed authentication issues

cooldogedev:


minecraft/protocol/login/data.go
```diff
@@ -198,7 +199,7 @@ type ClientData struct {
 	// and the username changed.
 	// Although this field is obviously here for a reason, allowing this is too dangerous and should never be
 	// done.
-	ThirdPartyNameOnly bool
+	ThirdPartyNameOnly *bool `json:"ThirdPartyNameOnly,omitempty"`
```
didntpot, line 202: you could just remove the field

cooldogedev: It's needed for backwards compatibility, Gophertunnel won't be able to dial servers using older protocols without it.

PR #328, opened 2025-07-04
Create a MultiplayerToken which can be used

Shadowsniper784:
I have a specific issue where I already have the XBL tokens from an oauth flow implemented externally, and so this PR hopes to allow me to use it in gophertunnel

### DIFF
--- a/minecraft/dial.go
+++ b/minecraft/dial.go
@@ -55,7 +55,7 @@ type Dialer struct {
 	TokenSource oauth2.TokenSource
 
 	// XBLToken should be used in place of TokenSource if the XBL token is already known, i.e through a different
-	// oauth source. This token is for with the https://multiplayer.minecraft.net relaying party
+	// oauth source. This token is for with the https://multiplayer.minecraft.net relaying party.
 	XBLToken *auth.XBLToken
 
 	// PacketFunc is called whenever a packet is read from or written to the connection returned when using


minecraft/dial.go
```diff
@@ -54,6 +54,10 @@ type Dialer struct {
 	// If TokenSource is nil, the connection will not use authentication.
 	TokenSource oauth2.TokenSource
 
+	// XBLToken should be used in place of TokenSource if the XBL token is already known, i.e through a different
+	// oauth source
+	MultiplayerToken *auth.XBLToken
```
Sandertv: Doc name and field name don't match. I think calling the field XBLToken like the comment describes is probably less confusing.

minecraft/dial.go
```diff
@@ -172,8 +176,15 @@ func (d Dialer) DialContext(ctx context.Context, network, address string) (conn
...
-	if d.TokenSource != nil {
-		chainData, err = authChain(ctx, d.TokenSource, key)
+	if d.TokenSource != nil || d.MultiplayerToken != nil {
+		xblToken, err := getXBLToken(d)
+		if err != nil {
+			return nil, &net.OpError{Op: "dial", Net: "minecraft", Err: err}
+		}
+		if xblToken == nil {
+			return nil, &net.OpError{Op: "dial", Net: "minecraft", Err: fmt.Errorf("no XBL token provided")}
+		}
```
Sandertv: This check should not be necessary.

minecraft/dial.go
```diff
@@ -340,21 +351,32 @@ func listenConn(conn *Conn, readyForLogin, connected chan struct{}, cancel conte
...
-// chain ready to be put in a login request is returned.
-func authChain(ctx context.Context, src oauth2.TokenSource, key *ecdsa.PrivateKey) (string, error) {
-	// Obtain the Live token, and using that the XSTS token.
-	liveToken, err := src.Token()
+func getXBLToken(dialer Dialer) (*auth.XBLToken, error) {
+	if dialer.MultiplayerToken != nil {
+		return dialer.MultiplayerToken, nil
+	}
+	if dialer.TokenSource == nil {
+		return nil, fmt.Errorf("no XBL token source provided")
```
Sandertv: This check is also unnecessary, since the block that calls this function already checks if either `dialer.MultiplayerToken` or `dialer.TokenSource` is not nil. When the code reaches this point, `dialer.TokenSource` is guaranteed not to be nil.

Sandertv:
Thanks for the PR! I've left a couple of comments, please feel free to let me know once you've resolved them.

minecraft/dial.go
```diff
@@ -172,8 +176,12 @@ func (d Dialer) DialContext(ctx context.Context, network, address string) (conn
 		return nil, &net.OpError{Op: "dial", Net: "minecraft", Err: fmt.Errorf("generating ECDSA key: %w", err)}
 	}
 	var chainData string
-	if d.TokenSource != nil {
-		chainData, err = authChain(ctx, d.TokenSource, key)
+	if d.TokenSource != nil || d.XBLToken != nil {
+		xblToken, err := getXBLToken(d)
```
HashimTheArab: pass the existing context

minecraft/dial.go
```diff
@@ -54,6 +54,10 @@ type Dialer struct {
 	// If TokenSource is nil, the connection will not use authentication.
 	TokenSource oauth2.TokenSource
 
+	// XBLToken should be used in place of TokenSource if the XBL token is already known, i.e through a different
+	// oauth source. This token is for with the https://multiplayer.minecraft.net relaying party
```
TwistedAsylumMC: Nitpick but could you just add a "." at the end of this doc?

PR #334, opened 2025-08-22
fix logic

FDUTCH:
made the code match the comment


TwistedAsylumMC:
The comment should be removed here, not the actual error. If you'd like to do that with this PR then I can merge

PR #336, opened 2025-08-24
Use microsoft server time in signed requests

HashimTheArab:
- Saves microsoft server time from responses and uses it in signed requests, many user devices may have out of sync time which will cause a 403 forbidden error from xbox api
- Changes "deviceToken" key to "DeviceToken" to match real request
- Simplifies w.Write to fmt.Fprintf

SImply saving serverTime to a variable works because Token() is always called during the dial process before sending the signed requests obtainXblToken and obtainDeviceToken

### DIFF
--- a/minecraft/auth/live.go
+++ b/minecraft/auth/live.go
@@ -100,30 +100,25 @@ func RequestLiveTokenWriter(w io.Writer) (*oauth2.Token, error) {
 }
 
 var (
-	serverDateMu sync.Mutex
-	// serverDate represents the most recent server date received from Microsoft servers.
+	serverTimeMu sync.Mutex
+	// serverTime represents the most recent server date received from Microsoft servers.
 	// It's used for the signed requests which can be blocked if the users device time is not synced.
 	// It uses the date received from the unsigned requests.
-	serverDate time.Time
+	serverTime time.Time
 )
 
-func getDateHeader(headers http.Header) time.Time {
+func updateServerTimeFromHeaders(headers http.Header) {
 	date := headers.Get("Date")
 	if date == "" {
-		return time.Time{}
+		return
 	}
-	if t, err := time.Parse(time.RFC1123, date); err == nil {
-		return t
+	t, err := time.Parse(time.RFC1123, date)
+	if err != nil || t.IsZero() {
+		return
 	}
-	return time.Time{}
-}
-
-func setServerDate(d time.Time) {
-	if !d.IsZero() {
-		serverDateMu.Lock()
-		serverDate = d
-		serverDateMu.Unlock()
-	}
+	serverTimeMu.Lock()
+	serverTime = t
+	serverTimeMu.Unlock()
 }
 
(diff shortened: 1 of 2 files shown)


minecraft/auth/live.go
```diff
@@ -97,6 +99,33 @@ func RequestLiveTokenWriter(w io.Writer) (*oauth2.Token, error) {
...
+}
+
+func setServerDate(d time.Time) {
+	if !d.IsZero() {
+		serverDateMu.Lock()
+		serverDate = d
+		serverDateMu.Unlock()
+	}
+}
+
```
TwistedAsylumMC: Since these two methods are only ever called together, I'd merge them into an `updateServerDate` method instead

minecraft/auth/live.go
```diff
@@ -97,6 +99,33 @@ func RequestLiveTokenWriter(w io.Writer) (*oauth2.Token, error) {
 	panic("unreachable")
 }
 
+var (
+	serverDateMu sync.Mutex
+	// serverDate represents the most recent server date received from Microsoft servers.
+	// It's used for the signed requests which can be blocked if the users device time is not synced.
+	// It uses the date received from the unsigned requests.
+	serverDate time.Time
```
TwistedAsylumMC: Bit of a nitpick, but would "server time" be a better way to describe this, since in Go it's called a time.Time compared to a Date in javascript etc.

minecraft/auth/live.go
```diff
@@ -128,13 +157,19 @@ func pollDeviceAuth(deviceCode string) (t *oauth2.Token, err error) {
...
 	poll := new(deviceAuthPoll)
 	if err := json.NewDecoder(resp.Body).Decode(poll); err != nil {
 		return nil, fmt.Errorf("POST https://login.live.com/oauth20_token.srf: json decode: %w", err)
 	}
-	if poll.Error == "authorization_pending" {
+	switch poll.Error {
+	case "authorization_pending":
 		return nil, nil
-	} else if poll.Error == "" {
+	case "":
```
TwistedAsylumMC, line 165: Curious as to why this was changed to a switch

minecraft/auth/xbox.go
```diff
@@ -72,7 +72,7 @@ func obtainXBLToken(ctx context.Context, c *http.Client, key *ecdsa.PrivateKey,
 	data, err := json.Marshal(map[string]any{
 		"AccessToken":       "t=" + liveToken.AccessToken,
 		"AppId":             "0000000048183522",
-		"deviceToken":       device.Token,
+		"DeviceToken":       device.Token,
```
TwistedAsylumMC, line 75: What's happened with this?

HashimTheArab: The correct one is capitalized matched it against many other implementations

minecraft/auth/live.go
HashimTheArab: VSC linter suggested it

PR #340, opened 2025-09-05
fix: flate decompression logic

HashimTheArab:
- Proper overflow detection for limit instead of just returning truncated data
- Fixed decompression pool usage, old logic put reader back in pool without closing in cases of Reset errors
- Old logic closed the reader before reading which worked but was wrong
- Improved capacity hints, does not set to * 2 if its not needed

### DIFF
--- a/minecraft/protocol/packet/compression.go
+++ b/minecraft/protocol/packet/compression.go
@@ -3,11 +3,13 @@ package packet
 import (
 	"bytes"
 	"fmt"
+	"io"
+	"math"
+	"sync"
+
 	"github.com/golang/snappy"
 	"github.com/klauspost/compress/flate"
 	"github.com/sandertv/gophertunnel/minecraft/internal"
-	"io"
-	"sync"
 )
 
 // Compression represents a compression algorithm that can compress and decompress data.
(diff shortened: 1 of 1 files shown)


minecraft/protocol/packet/compression.go
```diff
@@ -106,20 +108,43 @@ func (flateCompression) Compress(decompressed []byte) ([]byte, error) {
 
 // Decompress ...
 func (flateCompression) Decompress(compressed []byte, limit int) ([]byte, error) {
-	buf := bytes.NewReader(compressed)
-	c := flateDecompressPool.Get().(io.ReadCloser)
-	defer flateDecompressPool.Put(c)
+	r := flateDecompressPool.Get().(io.ReadCloser)
+	defer func() {
+		_ = r.Close()
```
Sandertv, line 110: I am not sure if we need to Close() at all. For zlib, the Close before reading was necessary for the checksum, but I believe deflate has no such checksum. Given we don't need it, maybe Close() isn't necessary at all?

minecraft/protocol/packet/compression.go
```diff
@@ -106,20 +108,43 @@ func (flateCompression) Compress(decompressed []byte) ([]byte, error) {
...
-	decompressed := bytes.NewBuffer(make([]byte, 0, len(compressed)*2))
-	if _, err := io.Copy(decompressed, io.LimitReader(c, int64(limit))); err != nil {
+	var decompressed bytes.Buffer
+	// If the compressed data is less than half the limit, we can safely assume l*2, otherwise cap at limit.
+	l := len(compressed)
+	capHint := limit
+	if l <= limit/2 {
+		capHint = l * 2
+	}
+	decompressed.Grow(capHint)
```
Sandertv: I believe this can be simplified to:
```go
decompressed := bytes.NewBuffer(make([]byte, 0, min(limit, len(compressed) * 2)))
```

minecraft/protocol/packet/compression.go
```diff
@@ -106,20 +108,43 @@ func (flateCompression) Compress(decompressed []byte) ([]byte, error) {
...
+
+	// Handle no limit
+	if limit == math.MaxInt {
+		if _, err := io.Copy(&decompressed, r); err != nil {
+			return nil, fmt.Errorf("decompress flate: %w", err)
+		}
+		return decompressed.Bytes(), nil
+	}
```
Sandertv, line 133: Is there ever a sensible case to have no limit? I get that this is probably to avoid overflow, but could just do a `max(limit + 1, math.MaxInt)` somewhere.

minecraft/protocol/packet/compression.go
```diff
@@ -106,20 +108,43 @@ func (flateCompression) Compress(decompressed []byte) ([]byte, error) {
...
+	// Read limit+1 bytes to detect overflow without CopyN truncating the result.
+	toRead := int64(limit) + 1
+	n, err := io.CopyN(&decompressed, r, toRead)
+	if err != nil && err != io.EOF {
+		// CopyN returns an EOF error if the stream is shorter than the limit, we can treat that as a success.
 		return nil, fmt.Errorf("decompress flate: %w", err)
 	}
+	if n > int64(limit) {
+		return nil, fmt.Errorf("decompress flate: size %d exceeds limit %d", n, limit)
+	}
```
Sandertv: Maybe a slightly easier to understand implementation, slightly closer to what it was before, might be this?
```go
lr := io.LimitReader(c, int64(limit) + 1)
if _, err := io.Copy(decompressed, lr); err != nil {
    return nil, fmt.Errorf("decompress flate: %w", err)
} else if lr.(*io.LimitedReader).N <= 0 {
    return nil, fmt.Errorf("decompress flate: size %d exceeds limit %d", n, limit)
}
```

minecraft/protocol/packet/compression.go
HashimTheArab: The if checking the l <= limit/2 protects against an int overflow on 32 bit devices if a packet is 1gb, which is impossible but having the if instead of min is technically correct

minecraft/protocol/packet/compression.go
HashimTheArab: Good idea

HashimTheArab: limit + 1 would cause an overflow so we cant handle this without checking limit == math.MaxInt, which is what the listener sets if the user configures no limit
https://github.com/Sandertv/gophertunnel/blob/8096f00d11d6d7ee3472b59fc443dfdf6911e29e/minecraft/listener.go#L133

Sandertv (approved):
I think these changes look sensible. I assume this has been tested thoroughly?

PR #342, opened 2025-09-14
fix: reading manifest fails if there are trailing comments

HashimTheArab:
jsonc.unmarshal only strips comments that are in the json block, not any that come after
image


TwistedAsylumMC:
Should be solved by https://github.com/Sandertv/gophertunnel/commit/f75ce5e7cfd51624acdefd373e90731872053c73

PR #343, opened 2025-09-15
Added missing chemistry packs

JavierLeon9966:
Source: BDS with EDU mode on


TwistedAsylumMC:
Just realised, should these packs not only be applied if edu features are enabled? If BDS only adds them to the stack when EDU is enabled then we should probably do the same

PR #345, opened 2025-09-16
Allow both semver and array in pack manifest versions and add missing Metadata field

### LINKED ISSUE #344: Resource pack manifest versions don't support semver
Every version field in the manifest could be either [3]int like [1, 0, 0] or a string ("1.0.0"). Currently gophertunnel doesn't allow strings.

# Source
https://github.com/Sandertv/gophertunnel/blob/cca6a7757bb0523a107fe2193b6a5e2899192eec/minecraft/resource/manifest.go#L38

# Docs
Vector [a, b, c] or [SemVer](https://semver.org/) String
https://learn.microsoft.com/en-us/minecraft/creator/reference/content/addonsreference/packmanifest?view=minecraft-bedrock-stable

# Error log

{"msg":"download resource pack: invalid full resource pack data: read manifest: decode manifest JSON: json: cannot unmarshal string into Go struct field Dependency.dependencies.version of type [3]int (data: {\n    \"format_version\": 2,\n    \"header\": {\n        \"name\": \"Better Water Sounds\",\n        \"description\": \"\",\n        \"uuid\": \"12475b6e-e839-4837-b774-a14207e3f618\",\n        \"version\": [\n            1,\n            0,\n            2\n        ],\n        \"min_engine_version\": [\n            1,\n            21,\n            70\n        ]\n    },\n    \"modules\": [\n        {\n            \"type\": \"resources\",\n            \"uuid\": \"9a6c056c-bc0d-4d22-a902-306bae15e692\",\n            \"version\": [\n                1,\n                0,\n                0\n            ]\n        }\n    ],\n    \"metadata\": {\n        \"authors\": [\n            \"jeanmajid\"\n        ]\n    },\n    \"dependencies\": [\n        {\n            \"uuid\": \"dd7d8871-272a-45d5-a07b-efba43b886fb\",\n            \"version\": \"1.0.2\"\n        }\n    ]\n})","UUID":"12475b6e-e839-4837-b774-a14207e3f618"}

HashimTheArab:
Closes #344 
image

### DIFF
--- a/minecraft/resource/manifest.go
+++ b/minecraft/resource/manifest.go
@@ -1,9 +1,54 @@
 package resource
 
-import "github.com/google/uuid"
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/google/uuid"
+)
 
 // Documentation on this may be found here:
 // https://learn.microsoft.com/en-us/minecraft/creator/reference/content/addonsreference/examples/addonmanifest
+
+// Version may be present in the manifest as [1,0,0] or "1.0.0".
+type Version [3]int
+
+func (v *Version) UnmarshalJSON(b []byte) error {
+	// Parse common array format [1,0,0]
+	var arr []int
+	if err := json.Unmarshal(b, &arr); err == nil && len(arr) > 0 {
+		for i := 0; i < 3 && i < len(arr); i++ {
+			v[i] = arr[i]
+		}
+		return nil
+	}
+
+	// Parse semver format "1.0.0"
+	var s string
+	if err := json.Unmarshal(b, &s); err == nil {
+		s = strings.TrimPrefix(strings.TrimSpace(s), "v")
+		if i := strings.IndexAny(s, "+-"); i >= 0 {
+			s = s[:i]
+		}
+
+		parts := strings.Split(s, ".")
+		for i := 0; i < 3 && i < len(parts); i++ {
+			if parts[i] == "" {
+				continue
+			}
+			n, err := strconv.Atoi(parts[i])
+			if err != nil {
+				return fmt.Errorf("invalid version component %q in %q", parts[i], s)
+			}
+			v[i] = n
+		}
+		return nil
+	}
+
+	return fmt.Errorf("invalid version: %s", string(b))
+}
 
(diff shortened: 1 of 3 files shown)


minecraft/resource/manifest.go
```diff
@@ -1,10 +1,55 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/google/uuid"
+)
 
 // Documentation on this may be found here:
 // https://learn.microsoft.com/en-us/minecraft/creator/reference/content/addonsreference/examples/addonmanifest
 
+// Version may be present in the manifest as [1,0,0] or "1.0.0".
```
TwistedAsylumMC: I'd move all this to be below the `Header` struct just to maintain order

minecraft/resource/manifest.go
```diff
@@ -1,10 +1,55 @@
...
+		return nil
+	}
+
+	// Parse semver format "1.0.0"
+	var s string
+	if err := json.Unmarshal(b, &s); err == nil {
+		s = strings.TrimPrefix(strings.TrimSpace(s), "v")
+		if i := strings.IndexAny(s, "+-"); i >= 0 {
+			s = s[:i]
+		}
```
TwistedAsylumMC: We can probably skip this since their official parsing/validator wouldn't allow for this
https://github.com/Mojang/minecraft-creator-tools/blob/main/app/src/app/Package.ts#L202-L217

minecraft/resource/manifest.go
```diff
@@ -1,10 +1,55 @@
...
+
+func (v *Version) UnmarshalJSON(b []byte) error {
+	// Parse common array format [1,0,0]
+	var arr []int
+	if err := json.Unmarshal(b, &arr); err == nil && len(arr) > 0 {
+		for i := 0; i < 3 && i < len(arr); i++ {
+			v[i] = arr[i]
+		}
+		return nil
+	}
```
TwistedAsylumMC: This can be simplified
```go
	var arr [3]int
	if err := json.Unmarshal(b, &arr); err == nil {
		*v = arr
		return nil
	}
```

TwistedAsylumMC (changes requested):
All looks good apart from a few things

PR #354, opened 2025-10-28
refactor: use klauspost/compress for snappy instead of archived golang/snappy

HashimTheArab:
no code changes are needed its a dropin replcaement and we already use the same lib for flate


Sandertv (approved):
Fair enough! Any ideas on why golang/snappy was archived?

PR #356, opened 2025-10-29
refactor: add conn.Disconnect and update listener.Disconnect to use that

HashimTheArab:


minecraft/conn.go
```diff
@@ -577,6 +577,17 @@ func (conn *Conn) Context() context.Context {
...
 }
 
+// Disconnect disconnects the connection by first sending a disconnect packet with the message passed, and
+// closing the connection after. If the message passed is empty, the client will be immediately sent to the
+// server list instead of a disconnect screen.
+func (conn *Conn) Disconnect(message string) error {
+	_ = conn.WritePacket(&packet.Disconnect{
+		HideDisconnectionScreen: message == "",
+		Message:                 message,
+	})
```
Sandertv, line 587: This sends a Disconnect packet to the server if it is a client connection. That is not a logical change.

PR #338, opened 2025-08-28
use new auth system when dialing, refactor auth

GameParrot:
- Support for new auth system when connecting to servers (TODO: support for offline logins and parsing the new token field)
- new session.Session for reusing tokens between logins
- new auth.XblTokenObtainer for reusing a device token
- Support for authenticating with platforms other than Android

PR #365, opened 2025-12-01
Auth

cjmustard:
This is the final version of @GameParrot auth PR:
https://github.com/Sandertv/gophertunnel/pull/338

Identical, but includes support for offline logins and parsing the new token field

### DIFF
--- a/minecraft/listener.go
+++ b/minecraft/listener.go
@@ -202,20 +202,25 @@ func Listen(network, address string) (*Listener, error) {
 
 // authEnvCache holds authorization environment used to issue or verify
 // the multiplayer token for OpenID authentication. It is cached by authEnv.
-var authEnvCache atomic.Pointer[service.AuthorizationEnvironment]
+var (
+	authEnvCache   *service.AuthorizationEnvironment
+	authEnvCacheMu sync.Mutex
+)
 
 // authEnv returns the authorization environment that can be used for issuing
 // or verifying the multiplayer token for OpenID authentication.
 // This method is only called once and cached globally which means it will
 // use the HTTP client from the first caller's context (oauth2.HTTPClient).
 func authEnv(ctx context.Context) (*service.AuthorizationEnvironment, error) {
-	if e := authEnvCache.Load(); e != nil {
-		return e, nil
+	authEnvCacheMu.Lock()
+	defer authEnvCacheMu.Unlock()
+	if authEnvCache != nil {
+		return authEnvCache, nil
 	}
+
 	if ctx == nil {
 		ctx = context.Background()
 	}
-
 	ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


cjmustard:
I believe this is all; it follows the same path as the PMMP auth implementation.

minecraft/franchise/internal/result.go
```diff
@@ -0,0 +1,11 @@
+package internal
+
+// Result wraps the basic structure of response body sent by most franchise-related services.
+// Make sure to specify the T generic type to whatever you want in the Data.
+type Result[T any] struct {
+	// Data is the payload of the result, if successful.
+	Data T `json:"result"`
+
+	// I'm not sure if there was an error result, but I've seen errors in HTTP status code with
+	// an empty response body rather than a body describing the error so this isn't the case.
```
TwistedAsylumMC: I'd remove this comment since it's not really contributing to anything. If we find that errors are sent here in the future, we can add that in later on

minecraft/franchise/discovery.go
```diff
@@ -0,0 +1,107 @@
...
+//	discovery, err := franchise.Discover(protocol.CurrentVersion)
+//	if err != nil {
+//		log.Fatalf("Error obtaining discovery: %s", err)
+//	}
+//
+//	// Look up and decode an environment for authorization.
+//	auth := new(franchise.AuthorizationEnvironment)
+//	if err := discovery.Environment(auth, franchise.EnvironmentTypeProduction); err != nil {
+//		log.Fatalf("Error reading environment for %q: %s", a.EnvironmentName(), err)
+//	}
```
TwistedAsylumMC: Would it be better to have an `ExampleDiscovery()` method (similar to `ExampleDial()` and `ExampleListen()`) so they show correctly in the go docs?

minecraft/franchise/multiplayer.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/sandertv/gophertunnel/minecraft/franchise/internal"
+)
+
+type MultiplayerToken struct {
```
TwistedAsylumMC: Can you document what a multiplayer token is, and how the user should use/create it (i.e. by  calling RequestMultiplayerToken)

TwistedAsylumMC (changes requested):
Just a few comments from looking through. As well as these, could you remove the token source stuff from `main.go`? We never included file storage for tokens because it isn't exactly a safe practice. If you could make it the minimal version needed to work, that would be great

HashimTheArab:
@codex review the pr

cjmustard:
bro got left on read by two ai models 😭

HashimTheArab:
resolving some issues and it will be ready for review soon. lactyy refactored a large part of this and made it much easier to review

HashimTheArab:
@GameParrot could you point out the spots in the code? I see `/.well-known/openid-configuration`, is there anything else? and how long of a cache would you recommend

GameParrot:
> @GameParrot could you point out the spots in the code? I see `/.well-known/openid-configuration`, is there anything else? and how long of a cache would you recommend

See https://github.com/pmmp/PocketMine-MP/blob/stable/src/network/mcpe/auth/AuthKeyProvider.php

didntpot (approved):

GameParrot:
Looking at the impl, it looks like if 2 listeners are created simultaneously it might create 2 authorization environments instead of using the cached one, unless I am missing something. Should either use a mutex on authEnv or create it on init.

PR #367, opened 2025-12-05
Implement compression threshold

Dasciam:


didntpot (approved):
LGTM!

NopeNotDark (approved):

AkmalFairuz (approved):
LGTM!

minecraft/listener.go
```diff
@@ -58,6 +58,10 @@ type ListenConfig struct {
 	// Compression is the packet.Compression to use for packets sent over this Conn. If set to nil, the compression
 	// will default to packet.flateCompression.
 	Compression packet.Compression // TODO: Change this to snappy once Windows crashes are resolved.
+	// CompressionThreshold specifies the minimum data size in bytes that triggers compression. Data smaller than this threshold
+	// will not be compressed. If zero, compression threshold will default to 1024.
```
Sandertv: Do we have any sort of benchmarks or theoretical basis to use 1024 as a default?

Sandertv (commented):
Seems like a good addition. Is the default limit of 1024 based on anything?

minecraft/listener.go
Dasciam: > Do we have any sort of benchmarks or theoretical basis to use 1024 as a default?

It's based on PM's `network.batch-threshold` option. It's 256 by default but I think 1024 would be better

Dasciam: > Do we have any sort of benchmarks or theoretical basis to use 1024 as a default?

If we say about benchmarks, I did many profiles of before and after, and noticed a 25% performance gain

Sandertv (approved):
I think this looks good and a compression threshold of 256 sounds reasonable. Thank you for the PR!

PR #368, opened 2025-12-09
minecraft/protocol: Support 1.21.130

cooldogedev:


JkqzDev (approved):

minecraft/protocol/packet/command_output.go
```diff
@@ -25,7 +25,7 @@ type CommandOutput struct {
 	CommandOrigin protocol.CommandOrigin
 	// OutputType specifies the type of output that is sent. The OutputType sent by vanilla games appears to
 	// be 3, which seems to work.
-	OutputType byte
+	OutputType string
```
Happy2018new: The comment of this field should also change.

```go
// OutputType specifies the type of output that is sent. The OutputType sent by vanilla games appears to
// be 3, which seems to work.
```

minecraft/protocol/packet/id.go
```diff
@@ -230,6 +230,7 @@ const (
 	IDClientBoundControlSchemeSet
 	IDDebugDrawer
 	IDServerBoundPackSettingChange
-	_
+	_ // TODO: Implement client-bound data store
```
Happy2018new, line 233: I think you can reference to another PR, and in that PR it implemented this packet I think.

minecraft/protocol/packet/text.go
```diff
@@ -57,8 +57,17 @@ func (*Text) ID() uint32 {
 }
 
 func (pk *Text) Marshal(io protocol.IO) {
-	io.Uint8(&pk.TextType)
 	io.Bool(&pk.NeedsTranslation)
+	var categoryType uint8
+	if pk.TextType == TextTypeRaw || pk.TextType == TextTypeTip || pk.TextType == TextTypeSystem || pk.TextType == TextTypeObjectWhisper || pk.TextType == TextTypeObjectAnnouncement || pk.TextType == TextTypeObject {
```
Happy2018new: You better change to something like that

```go
switch pk.TextType {
    case TextTypeRaw, TextTypeTip, TextTypeSystem, TextTypeObjectWhisper, TextTypeObjectAnnouncement, TextTypeObject:
        categoryType = protocol.TextCategoryMessageOnly
    case TextTypeChat, TextTypeWhisper, TextTypeAnnoument:
        categoryType = protocol.TextCategoryAuthoredMessage
    default:
        categoryType = protocol.TextCategoryMessageWithParameters
}
```

minecraft/protocol/reader.go
```diff
@@ -642,6 +643,25 @@ func (r *Reader) ShapeData(x *ShapeData) {
 	(*x).Marshal(r)
 }
 
+// TextCategory reads a text category from the reader.
+func (r *Reader) TextCategory(x *uint8) {
```
Happy2018new: This function make me confused if I not see the implementation in `protocol.Writer`. I think it is better to store the string but not discard them.

minecraft/protocol/packet/animate.go
```diff
@@ -5,32 +5,40 @@ import (
 )
 
 const (
-	AnimateActionSwingArm = iota + 1
+	AnimateActionSwingArm = uint8(iota + 1)
```
AkmalFairuz: This is more idiomatic Go than using `uint8(iota + 1)`

```go
    AnimateActionSwingArm uint8 = iota + 1
```

minecraft/protocol/text.go
```diff
@@ -0,0 +1,13 @@
+package protocol
+
+const (
+	TextCategoryMessageOnly = uint8(iota)
```
AkmalFairuz: This is more idiomatic Go than using `uint8(iota)`.

```go
    TextCategoryMessageOnly uint8 = iota
```

minecraft/protocol/packet/animate.go
cooldogedev: More idiomatic by what standard? https://github.com/golang/net/blob/master/quic/errors.go#L19-L38

minecraft/protocol/reader.go
cooldogedev: The user doesn't need to know the category types, they can extract such information from the message type directly.

minecraft/protocol/packet/id.go
cooldogedev: I'm not linking a draft here, I've left a todo, which clearly gets the point across.

minecraft/protocol/packet/animate.go
AkmalFairuz: https://tip.golang.org/wiki/Iota

minecraft/protocol/reader.go
Happy2018new: OK, I see that. Maybe you could change the internal implements to make this function more readable, because those who need maintain gophertunnel may get confused. (No need to store the string, just find another way to make the internal implements more readable)

PR #373, opened 2025-12-23
mineraft/protocol: Simple event packet can send from client

Happy2018new:
When change the default game type to creative mode, this packet will send from mc to the server.
```go
packet.SimpleEvent{
    EventType: packet.SimpleEventCommandsEnabled
}
```
However, the client packet pool not include this packet, then the connection between mc and server will lost due to unknown packet error.

-----

I fix this and change the description of this packet.
The description is the same to mojang's docs.
> image

-----

However, I don't know when is the time that the server send this packet.
If the server will never send that, then this packet should remove from the server packet pool.


didntpot (approved):

Sandertv (approved):
Looks good. Thank you for the PR!

PR #376, opened 2025-12-26
perf(nbt): avoid reflect SetMapIndex for map[string]any compounds

HashimTheArab:
- Addresses Constant reflect.Ptr should be inlined [gofix](https://pkg.go.dev/golang.org/x/tools/internal/gofix)
- Optimize NBT decoding which is using more memory than chunks. All usage is from SetMapIndex calls.

The old path for TAG_Compound -> map[string]any used reflection (reflect.MakeMap + SetMapIndex), which is allocation-heavy and creates lots of short‑lived reflect.Value/interface churn.

## Before
Total program usage is 400mb, 26% is from nbt decoding
image

## After
This is a smaller sample size (155 mb vs 450 mb) but memory usage from NBT decoding has basically dropped to non existent. 26% of total usage before, and now 4%.
image

### DIFF
--- a/minecraft/nbt/decode.go
+++ b/minecraft/nbt/decode.go
@@ -431,41 +431,41 @@ func (d *Decoder) unmarshalTag(val reflect.Value, t tagType, tagName string) err
 			}
 			fieldMapPool.Put(fields)
 		case reflect.Interface, reflect.Map:
-			if vk := val.Kind(); vk == reflect.Interface && val.NumMethod() != 0 {
-				return InvalidTypeError{Off: d.r.off, FieldType: val.Type(), Field: tagName, TagType: t}
+			valKind := val.Kind()
+			valType := val.Type()
+			if valKind == reflect.Interface && val.NumMethod() != 0 {
+				return InvalidTypeError{Off: d.r.off, FieldType: valType, Field: tagName, TagType: t}
 			}
 
 			// Fast path for TAG_Compound into map[string]any and any(map[string]any).
 			// Avoid reflect.Map SetMapIndex; reuse+clear maps and fill via native assignments.
-			if val.Kind() == reflect.Map && val.Type() == mapStringAnyType {
-				m := val.Interface().(map[string]any)
-				if m == nil {
+			if (valKind == reflect.Map && valType == mapStringAnyType) || (valKind == reflect.Interface && isAny(val)) {
+				var m map[string]any
+				if valKind == reflect.Map {
+					m = val.Interface().(map[string]any)
+				} else {
+					m, _ = val.Interface().(map[string]any)
+				}
+
+				allocated := m == nil
+				if allocated {
 					m = make(map[string]any)
+				} else {
+					clear(m)
+				}
+
+				// Needed when decoding into an interface slot, or when the map was nil and we created a new one.
+				if valKind == reflect.Interface || allocated {
 					val.Set(reflect.ValueOf(m))
-				} else {
-					clear(m)
 				}
-				if err := d.unmarshalCompoundAny(m); err != nil {
-					return err
-				}
-				break
-			}
-			if val.Kind() == reflect.Interface && isAny(val) {
-				m, _ := val.Interface().(map[string]any)
-				if m == nil {
-					m = make(map[string]any)
-				} else {
-					clear(m)
-				}
-				val.Set(reflect.ValueOf(m))
+
 				if err := d.unmarshalCompoundAny(m); err != nil {
 					return err
 				}
 				break
 			}
 
-			valType := val.Type()
-			if val.Kind() == reflect.Map {
+			if valKind == reflect.Map {
 				valType = valType.Elem()
 			}
 			m := reflect.MakeMap(reflect.MapOf(stringType, valType))


HashimTheArab:
Ran some benchmarks
Before
```
BenchmarkDecoderDecode_Compound_MapStringAny_NewMap-11             29000             39244 ns/op           48544 B/op        836 allocs/op
BenchmarkDecoderDecode_Compound_MapStringAny_ReusedMap-11          34944             33763 ns/op           48536 B/op        835 allocs/op
BenchmarkDecoderDecode_Compound_Any_New-11                         35120             33819 ns/op           48552 B/op        836 allocs/op
BenchmarkDecoderDecode_Compound_Any_Reused-11                      35148             37043 ns/op           48536 B/op        835 allocs/op
BenchmarkDecoderDecode_Compound_TypedMap_Fallback-11               22407             52878 ns/op           66849 B/op       1558 allocs/op
```

After
```
BenchmarkDecoderDecode_Compound_MapStringAny_NewMap-11             45678             29050 ns/op           40080 B/op        307 allocs/op
BenchmarkDecoderDecode_Compound_MapStringAny_ReusedMap-11          76930             15817 ns/op            2736 B/op        291 allocs/op
BenchmarkDecoderDecode_Compound_Any_New-11                         48909             24400 ns/op           40088 B/op        307 allocs/op
BenchmarkDecoderDecode_Compound_Any_Reused-11                      75871             15315 ns/op            2736 B/op        291 allocs/op
BenchmarkDecoderDecode_Compound_TypedMap_Fallback-11               22616             53057 ns/op           66848 B/op       1558 allocs/op
```

Benchstat results (10 samples per benchmark)
image

Benchstat summary
image


# Code
```go
package nbt

import (
	"bytes"
	"fmt"
	"testing"
)

var (
	benchCompoundData = func() []byte {
		root := make(map[string]any, 512)
		for i := 0; i < 256; i++ {
			root[fmt.Sprintf("k%d", i)] = int32(i)
		}
		root["flag"] = true // Encodes as TAG_Byte.
		root["nested"] = map[string]any{
			"a": int32(1),
			"b": int32(2),
			"c": "hello",
			"d": []any{int32(1), int32(2), int32(3), int32(4), int32(5)},
		}
		root["nested2"] = map[string]any{
			"x": map[string]any{
				"y": map[string]any{
					"z": int32(999),
				},
			},
		}
		root["list_any_i32"] = []any{int32(10), int32(11), int32(12), int32(13), int32(14)}

		b, err := Marshal(root)
		if err != nil {
			panic(err)
		}
		return b
	}()

	benchInt32MapData = func() []byte {
		root := make(map[string]int32, 512)
		for i := 0; i < 512; i++ {
			root[fmt.Sprintf("k%d", i)] = int32(i)
		}
		b, err := Marshal(root)
		if err != nil {
			panic(err)
		}
		return b
	}()

	benchSinkAny any
	benchSinkInt int
)

func BenchmarkDecoderDecode_Compound_MapStringAny_NewMap(b *testing.B) {
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		dec := NewDecoder(bytes.NewReader(benchCompoundData))
		var m map[string]any
		if err := dec.Decode(&m); err != nil {
			b.Fatal(err)
		}
		benchSinkInt += len(m)
		benchSinkAny = m["nested"]
	}
}

func BenchmarkDecoderDecode_Compound_MapStringAny_ReusedMap(b *testing.B) {
	b.ReportAllocs()
	m := map[string]any{"stale": int32(1)}
	for i := 0; i < b.N; i++ {
		dec := NewDecoder(bytes.NewReader(benchCompoundData))
		if err := dec.Decode(&m); err != nil {
			b.Fatal(err)
		}
		benchSinkInt += len(m)
		benchSinkAny = m["nested"]
	}
}

func BenchmarkDecoderDecode_Compound_Any_New(b *testing.B) {
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		dec := NewDecoder(bytes.NewReader(benchCompoundData))
		var a any
		if err := dec.Decode(&a); err != nil {
			b.Fatal(err)
		}
		m := a.(map[string]any)
		benchSinkInt += len(m)
		benchSinkAny = m["nested"]
	}
}

func BenchmarkDecoderDecode_Compound_Any_Reused(b *testing.B) {
	b.ReportAllocs()
	var a any = map[string]any{"stale": int32(1)}
	for i := 0; i < b.N; i++ {
		dec := NewDecoder(bytes.NewReader(benchCompoundData))
		if err := dec.Decode(&a); err != nil {
			b.Fatal(err)
		}
		m := a.(map[string]any)
		benchSinkInt += len(m)
		benchSinkAny = m["nested"]
	}
}

func BenchmarkDecoderDecode_Compound_TypedMap_Fallback(b *testing.B) {
	// This should keep using the reflection fallback (SetMapIndex) because it's not map[string]any.
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		dec := NewDecoder(bytes.NewReader(benchInt32MapData))
		var m map[string]int32
		if err := dec.Decode(&m); err != nil {
			b.Fatal(err)
		}
		benchSinkInt += len(m)
	}
}
```

minecraft/nbt/decode.go
```diff
@@ -397,6 +434,36 @@ func (d *Decoder) unmarshalTag(val reflect.Value, t tagType, tagName string) err
...
+				if m == nil {
+					m = make(map[string]any)
+				} else {
+					clear(m)
+				}
+				val.Set(reflect.ValueOf(m))
+				if err := d.unmarshalCompoundAny(m); err != nil {
+					return err
+				}
+				break
```
Sandertv: These two blocks are largely identical. Is there any way we could re-use the code?

Sandertv (commented):
It's not pretty, but then neither is the rest of the nbt package. A compromise worth making looking at the benchmarks you've posted.
I've left one comment to see if we can re-use some code. Looks good otherwise.

minecraft/nbt/decode.go
HashimTheArab: In the first block we have `m := val.Interface().(map[string]any)`, in the second its `m, _ := val.Interface().(map[string]any)`

this is because the first one is checking if its a reflect.Map and asserting, its meant to be guaranteed safe so we should let it panic if something is wrong

the 2nd one is using reflect.Interface to check `any(map[string]any)` and asserting, without the check a nil any type would panic here

also different cases for calling val.Set

they're handling 2 slightly different cases with not all the same logic so i think its easier to understand if they stay separated but to reduce duplicate code it would look like this
```go
if (valKind == reflect.Map && valType == mapStringAnyType) || (valKind == reflect.Interface && isAny(val)) {
	var m map[string]any
	if valKind == reflect.Map {
		m = val.Interface().(map[string]any)
	} else {
		m, _ = val.Interface().(map[string]any)
	}

	created := m == nil
	if created {
		m = make(map[string]any)
	} else {
		clear(m)
	}

	if valKind == reflect.Interface || created {
		val.Set(reflect.ValueOf(m))
	}

	if err := d.unmarshalCompoundAny(m); err != nil {
		return err
	}
	break
}
```

more compact but harder to understand since its mixing different logic into 1 block. im fine with either approach

HashimTheArab: leaning towards this the more i look at it lol lmk

HashimTheArab:
Finished requested changes

Sandertv (approved):
Looks good, thank you for the PR!

PR #377, opened 2025-12-26
perf: reduce Item and ItemInstance decode allocations

HashimTheArab:
- Avoid allocating fresh map[string]any for empty/absent item NBT
- Clear stale ItemInstance.StackNetworkID when hasNetID is false
- UUID decode uses io.ReadFull and avoids heap allocations. This one is a micro opt but imo worth keeping

Makes a meaningful difference to reduce allocations in CraftingData decode

```
// ReadFull reads exactly len(buf) bytes from r into buf.
// It returns the number of bytes copied and an error if fewer bytes were read.
// The error is EOF only if no bytes were read.
// If an EOF happens after reading some but not all the bytes,
// ReadFull returns [ErrUnexpectedEOF].
// On return, n == len(buf) if and only if err == nil.
// If r returns an error having read at least len(buf) bytes, the error is dropped.
```

## Benchstat summary
image

## Benchstat results
image

### DIFF
--- a/minecraft/protocol/reader.go
+++ b/minecraft/protocol/reader.go
@@ -427,10 +427,8 @@ func (r *Reader) ItemInstance(i *ItemInstance) {
 	if x.NetworkID == 0 {
 		// The item was air, so there is no more data we should read for the item instance. After all, air
 		// items aren't really anything.
-		x.MetadataValue, x.Count, x.BlockRuntimeID = 0, 0, 0
-		x.NBTData = nil
-		x.CanBePlacedOn, x.CanBreak = nil, nil
-		i.StackNetworkID = 0
+		x.MetadataValue, x.Count, x.BlockRuntimeID, i.StackNetworkID = 0, 0, 0, 0
+		x.NBTData, x.CanBePlacedOn, x.CanBreak = nil, nil, nil
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


minecraft/protocol/reader.go
```diff
@@ -423,12 +423,14 @@ func (r *Reader) ItemDescriptorCount(i *ItemDescriptorCount) {
...
-	x.NBTData = make(map[string]any)
 	r.Varint32(&x.NetworkID)
 	if x.NetworkID == 0 {
 		// The item was air, so there is no more data we should read for the item instance. After all, air
 		// items aren't really anything.
-		x.MetadataValue, x.Count, x.CanBePlacedOn, x.CanBreak = 0, 0, nil, nil
+		x.MetadataValue, x.Count, x.BlockRuntimeID = 0, 0, 0
+		x.NBTData = nil
+		x.CanBePlacedOn, x.CanBreak = nil, nil
+		i.StackNetworkID = 0
```
Sandertv: Any reason not to merge these like they were before?

HashimTheArab: I can change it it just became hard to read

Sandertv: If you just group the zeros and the nils that'd be good I think

Sandertv (approved):

PR #378, opened 2026-01-03
data.go: introduced DeviceID formats.

TrippleAWap:
Introduces validation for string fields, array sizes, and device IDs in ClientData to prevent malformed or malicious input. Device ID validation now supports multiple formats based on device model, including GUID, Base64, and BigInt hex. These checks help prevent server crashes and ensure data integrity.

### DIFF
--- a/minecraft/protocol/os.go
+++ b/minecraft/protocol/os.go
@@ -2,7 +2,7 @@ package protocol
 
 // DeviceOS is a device DeviceOS identifier. It holds a value of one of the constants below and may be found
 // in packets such as the Login packet.
-type DeviceOS int32
+type DeviceOS int
 
 const (
 	DeviceAndroid DeviceOS = iota + 1


Sandertv:
Definitely not opposed to more documentation and checks on the device ID. I think the introduction of DeviceIDFormat is a good idea.

TrippleAWap:
I don't know if we should have the DeviceIDFormat check or if we should let the user decide to implement it if they want. If you have any thoughts on this please let me know!

minecraft/protocol/login/data.go
```diff
@@ -306,7 +307,7 @@ var checkVersion = regexp.MustCompile("[0-9.]").MatchString
 
 // Validate validates the client data. It returns an error if any of the fields checked did not carry a valid
 // value.
-func (data ClientData) Validate() error {
+func (data *ClientData) Validate() error {
```
TwistedAsylumMC: Why was this made a pointer?

minecraft/protocol/login/data.go
```diff
@@ -387,3 +388,55 @@ func base64DecLength(base64Data string, validLengths ...int) error {
...
+// DeviceIDFormat describes the format of the DeviceID,
+// the documentation for each format is as follows:
+//
+//	DeviceOS to DeviceIDFormat
+//		DeviceAndroid: DeviceIDFormatLowerHexString, // "a4f365bb1e04459bbe4cb3cbf9d546e0",
+//		DeviceIOS:  DeviceIDFormatUpperHexString, // "ADA3DFA4622F4E2FB2C14A496D52DB96",
+//		DeviceWin32: DeviceIDFormatLowerHexString, // "362b850f82493ec2719ea267dd25167f",
+//		DeviceOrbis: DeviceIDFormatUUID, // "05601fd2-9c71-30b1-b174-5fa11b9de09f",
+//		DeviceXBOX: DeviceIDFormatBase64, // "VlhnpI7TuWyfHiUx3WYwFvQQHbDkv505h6VVo40Cngw=",
+type DeviceIDFormat uint8
```
TwistedAsylumMC, line 418: Does this formatting stay after `go fmt`? Just curious

minecraft/protocol/login/data.go
```diff
@@ -387,3 +388,55 @@ func base64DecLength(base64Data string, validLengths ...int) error {
...
+
+const (
+	DeviceIDFormatUpperHexString DeviceIDFormat = iota
+	DeviceIDFormatLowerHexString
+	DeviceIDFormatBase64
+	DeviceIDFormatUUID
+	DeviceIDFormatInvalid
+)
+
+func (dId DeviceID) Format() DeviceIDFormat {
```
TwistedAsylumMC: Just put a basic doc saying it tries to work out what format the device id uses and returns invalid if unknown

minecraft/protocol/login/data.go
```diff
@@ -387,3 +388,55 @@ func base64DecLength(base64Data string, validLengths ...int) error {
...
+	deviceId := string(dId)
+
+	// This must be checked before UUID as these are valid UUIDs as well.
+	data, err := hex.DecodeString(deviceId)
+	if err == nil {
+		if len(data) != 32 {
+			return DeviceIDFormatInvalid
+		}
+
+		lowerMatch := regexp.MustCompile("[a-z0-9]")
```
TwistedAsylumMC: You should probably precompile this regex so it's not created every time

minecraft/protocol/packet/add_player.go
```diff
@@ -58,10 +59,10 @@ type AddPlayer struct {
 	EntityLinks []protocol.EntityLink
 	// DeviceID is the device ID set in one of the files found in the storage of the device of the player. It
 	// may be changed freely, so it should not be relied on for anything.
-	DeviceID string
+	DeviceID login.DeviceID
 	// BuildPlatform is the build platform/device OS of the player that is about to be added, as it sent in
 	// the Login packet when joining.
-	BuildPlatform int32
+	BuildPlatform protocol.DeviceOS
```
TwistedAsylumMC: Do these need to change here? I think it's fine to cast them into their primitive types when passing them into the packets, just because it makes the Marshal slightly ugly. I don't believe we do this elsewhere, but please correct me if I'm wrong

minecraft/dial.go
```diff
@@ -403,7 +403,8 @@ func defaultClientData(address, username string, d *login.ClientData) {
 		d.ClientRandomID = rand.Int63()
 	}
 	if d.DeviceID == "" {
-		d.DeviceID = uuid.New().String()
+		// this can be parsed as a bigint and is standard on `most` devices
+		d.DeviceID = login.DeviceID(strings.ReplaceAll(uuid.NewString(), "-", ""))
```
TwistedAsylumMC, line 445: What device id format are we trying to replicate here? Surely this would return invalid with its own logic?

minecraft/protocol/login/data.go
TrippleAWap: I haven't tested this and correct me if I'm wrong but based on prior experience (in other languages) if we take it by a pointer it removes the expensive allocations of copying it and I don't see a reason as to not change it to a pointer. This usually wouldn't make much of a difference but ClientData is a rather heavy packet and if the behavior is as expected it copies the entire struct on call which is rather unnecessary.

minecraft/protocol/login/data.go
TrippleAWap: Yeah, just tested and it stays after `go fmt`

minecraft/protocol/login/data.go
TrippleAWap: Sounds good

minecraft/protocol/login/data.go
TrippleAWap: Will do.

minecraft/protocol/packet/add_player.go
TrippleAWap: No they aren't, it just makes it easier if you're handling the data but it's not that big of an issue; we can revert it.

minecraft/dial.go
TrippleAWap: DeviceIDFormatLowerHexString

TrippleAWap: Forgot to change the doc on that line, oops.

TrippleAWap: I'll also change the logic for generating it as it's rather unclear.

minecraft/protocol/packet/add_player.go
TwistedAsylumMC: Yeah I think that would be better. You can add in the docs that they're normally of a certain type etc

minecraft/dial.go
TwistedAsylumMC: Ah in my head I forgot uuids were 36 in length, I think with a better doc this line is fine

TrippleAWap: I'm actually going to keep the UUID to generate the hex string as it's much longer using rand.Read and it requires an extra error check if we were to use that. However, I wrote tests for all logic and fixed 2 issues; everything is tested and working in the latest commit.

minecraft/protocol/login/data.go
TrippleAWap: > Why was this made a pointer?

If we want this reverted I will do so, I don't know if Go copies it for sure but from prior experiences in other languages I would believe that'd be the behavior.

TwistedAsylumMC: Yeah Go does copy when used as a value, it's not the *end* of the world but I'm also happy for you to leave this change in

TrippleAWap: Alright sounds good! :)

PR #383, opened 2026-01-09
feat: added PropertyData to GameData

Josscoder:


minecraft/game_data.go
```diff
@@ -110,4 +110,9 @@ type GameData struct {
 	// its index in the expected block palette. This is useful for servers that wish to support multiple protocol versions
 	// and custom blocks, but it will result in extra bytes being written for every block in a sub chunk palette.
 	UseBlockNetworkIDHashes bool
+	// PropertyData contains properties that should be applied on the player. These properties are the same as the
+	// ones that are sent in the SyncActorProperty packet. This allows you to set custom properties on players,
+	// such as minecraft:health or custom properties like minecraft:is_dyno, which can be used by resource packs
```
TwistedAsylumMC: I wouldn't use minecraft prefixed properties as an example of custom properties, just do like `gophertunnel:example` or something

PR #410, opened 2026-03-28
Fix DDUI related packet

Azvyl:
Fix ServerBoundDataDrivenScreenClosed, DataStoreChangeEntry.ChangeType, and DataStoreUpdate.ControlType

This changes reflect pmmp/bedrock-protocol implementation

### DIFF
--- a/minecraft/protocol/packet/serverbound_data_driven_screen_closed.go
+++ b/minecraft/protocol/packet/serverbound_data_driven_screen_closed.go
@@ -4,20 +4,12 @@ import (
 	"github.com/sandertv/gophertunnel/minecraft/protocol"
 )
 
-const (
-	DataDrivenScreenCloseReasonProgrammaticClose = iota
-	DataDrivenScreenCloseReasonProgrammaticCloseAll
-	DataDrivenScreenCloseReasonClientCanceled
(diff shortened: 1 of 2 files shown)


minecraft/protocol/packet/serverbound_data_driven_screen_closed.go
```diff
@@ -4,20 +4,12 @@ import (
...
 // ServerBoundDataDrivenScreenClosed is sent by the client when a data-driven UI screen is closed.
 type ServerBoundDataDrivenScreenClosed struct {
-	// FormID is the optional unique instance ID of the form that was closed.
-	FormID protocol.Optional[uint32]
-	// CloseReason is the reason the screen was closed. It is one of the DataDrivenScreenCloseReason constants.
-	CloseReason uint8
+	// FormID is the unique instance ID of the form that was closed.
+	FormID uint32
+	// CloseReason is the reason the screen was closed.
+	CloseReason string
```
TwistedAsylumMC, line 12: This is still supposed to be a uint8 but is encoded as a string. Look at the other cases (I believe AnimateActor packet is one example) that does the string mapping

TwistedAsylumMC (changes requested):

PR #412, opened 2026-03-29
protocol/shape.go: Add uint8 type to DebugDrawerShape constants

hydzilla:
Added `uint8` type to `DebugDrawerShape` constants to match the underlying type used by the shape.


TwistedAsylumMC:
I forgot to merge this before the protocol update, so I have copied this change myself since there's conflicts after renaming everything. Thank you for the PR though!

PR #413, opened 2026-04-01
minecraft/protocol: Use EntityMetadata type instead of map[uint32]any

hydzilla:
Uses the `EntityMetadata` type instead of `map[uint32]any` in packets and IO methods. This allows helper methods like `SetFlag` and `UnsetFlag` to be used directly on packet fields without a type conversion.


hydzilla:
Could this be merged?

HashimTheArab:
lgtm

PR #420, opened 2026-04-26
Persist BlockingTick on shield ItemStacks

### LINKED ISSUE #304: Improper State Changing when Crouching with a item in Offhand
Currently when you whole a shield (or any other item) in your offhand, then crouch. Once you uncrouch and check your inventory you will notice that you are no longer able to remove the shield from your offhand slot.

HashimTheArab:
## Summary
  - Adds a `BlockingTick int64` field to `protocol.ItemStack` so the shield's blocking-tick value is preserved
  across decode/encode instead of being read-and-discarded / written-as-zero.
  - Wires the new field through both `Item` and `ItemInstance` paths in `reader.go` and `writer.go`.

  Fixes #304. When a proxy round-trips an `InventoryContent` for the off-hand slot, the shield's `blockingTick` was
   being silently zeroed, which left the client stuck in the blocking pose / desynced raise animation. Storing the
  field on `ItemStack` is the minimal fix suggested by @Happy2018new and confirmed by @JAMadison in the issue
  thread.

### DIFF
--- a/minecraft/protocol/item.go
+++ b/minecraft/protocol/item.go
@@ -36,6 +36,8 @@ type ItemStack struct {
 	CanBreak []string
 	// HasNetworkID ...
 	HasNetworkID bool
+	// BlockingTick is the tick at which a shield started blocking. It is only used for shield items.
+	BlockingTick int64
 }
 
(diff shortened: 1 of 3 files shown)


hydzilla:
Was this added in 1.26.10 or was this always incorrect?

PR #427, opened 2026-05-11
Reject malformed oversized packet batches during decode

cqdetdev:
## Summary

  - Reject empty inner packets while splitting a batch.
  - Reject truncated inner packet lengths before reading packet data.
  - Enforce the max packet count during batch splitting instead of after collecting the full slice.
  - Reject compressed batches that are missing the compression algorithm byte.
  - Preserve the batch header in the generic `io.Reader` decoder path.

  This prevents malformed compressed batches from forcing the decoder to allocate a large packet slice before returning
  an error.

## Testing
A vanilla dragonfly server and a patched dragonfly server (using this commit) were both tested with exploit. The vanilla server increased in RAM by 3gb after 16777216 packets were sent. The patched one did not. Both servers kicked the user.

### DIFF
--- a/minecraft/protocol/packet/decoder.go
+++ b/minecraft/protocol/packet/decoder.go
@@ -66,6 +66,7 @@ func NewDecoder(reader io.Reader) *Decoder {
 	return &Decoder{
 		r:                 reader,
 		buf:               make([]byte, 1024*1024*3),
+		header:            batch,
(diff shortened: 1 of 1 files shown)


HashimTheArab (approved):
LGTM, before this a malicious batch could just fill with millions of tiny packets and the decoder keeps growing the packets slice

PR #429, opened 2026-05-15
Remove scalar reads/writes from heap allocations

cqdetdev:
## Summary

This removes per-call heap allocations from fixed-width scalar reads and writes in the protocol and NBT encoders/decoders.

Instead of allocating fresh 2/4/8 byte slices for each scalar operation, `Reader`, `Writer`, `offsetReader`, and `offsetWriter` now reuse small 8-byte scratch buffers. Variable-length strings and slices are unchanged.

## Performance

Median results:

  | Benchmark | Before | After | Delta |
  | --- | ---: | ---: | ---: |
  | protocol read scalars | 488.0 ns/op, 48 B/op, 9 allocs/op | 123.6 ns/op, 0 B/op, 0 allocs/op | -74.7% |
  | protocol write scalars | 440.7 ns/op, 48 B/op, 9 allocs/op | 117.8 ns/op, 0 B/op, 0 allocs/op | -73.3% |
  | NBT little-endian read scalars | 337.1 ns/op, 32 B/op, 5 allocs/op | 132.1 ns/op, 0 B/op, 0 allocs/op | -60.8% |
  | NBT little-endian write scalars | 281.5 ns/op, 32 B/op, 5 allocs/op | 89.6 ns/op, 0 B/op, 0 allocs/op | -68.2% |
  | NBT big-endian read scalars | 329.3 ns/op, 32 B/op, 5 allocs/op | 126.3 ns/op, 0 B/op, 0 allocs/op | -61.6% |
  | NBT big-endian write scalars | 282.6 ns/op, 32 B/op, 5 allocs/op | 88.5 ns/op, 0 B/op, 0 allocs/op | -68.7% 

Long-term results will be lower allocations -> less GC pressure & less CPU allocation on decoding

### DIFF
--- a/minecraft/nbt/encoding_big_endian.go
+++ b/minecraft/nbt/encoding_big_endian.go
@@ -12,8 +12,9 @@ type bigEndian struct{}
 
 // WriteInt16 ...
 func (bigEndian) WriteInt16(w *offsetWriter, x int16) error {
-	b := *(*[2]byte)(unsafe.Pointer(&x))
-	if _, err := w.Write(b[:]); err != nil {
+	b := w.buf[:2]
+	binary.BigEndian.PutUint16(b, uint16(x))
+	if _, err := w.Write(b); err != nil {
 		return FailedWriteError{Op: "WriteInt16", Off: w.off}
 	}
 	return nil
@@ -21,8 +22,9 @@ func (bigEndian) WriteInt16(w *offsetWriter, x int16) error {
 
 // WriteInt32 ...
 func (bigEndian) WriteInt32(w *offsetWriter, x int32) error {
-	b := *(*[4]byte)(unsafe.Pointer(&x))
-	if _, err := w.Write(b[:]); err != nil {
+	b := w.buf[:4]
+	binary.BigEndian.PutUint32(b, uint32(x))
+	if _, err := w.Write(b); err != nil {
 		return FailedWriteError{Op: "WriteInt32", Off: w.off}
 	}
 	return nil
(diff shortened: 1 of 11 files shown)


HashimTheArab:
LGTM, the scalar buffer is the big improvement, I use that pattern in many places on lunar. Verified thoroughly with codex that the changes from unsafe to BE/LE directly should not have any difference on behavior and it ran many tests to make sure. I'm going to run this on prod.

> The scratch buffer change is the real allocation win here. The old scalar paths either allocated small []byte values for reads/writes or copied values into temporary arrays before writing. Reusing an [8]byte buffer on the reader/writer removes those per-scalar heap allocations while keeping the encoded bytes stable.

> Replacing scalar unsafe casts with encoding/binary should not change encoding. The previous unsafe code relied on native byte order: on little-endian builds it reinterpreted values as little-endian bytes, and on big-endian builds the build tags selected the big-endian implementation. The new code makes that same byte order explicit with binary.LittleEndian or binary.BigEndian. Signed integers are converted through their unsigned representation, which preserves the same two’s-complement bit pattern. Floats use math.Float32bits/math.Float64bits, which preserves the exact IEEE-754 bit pattern, including signed zero, infinities, and NaN payloads.

> I also checked representative edge cases locally: min/max signed integers, -1, zero, positive values, +0,
  -0, infinities, and custom NaN bit patterns. The bytes produced by the old unsafe representation matched
  the bytes produced by encoding/binary on the corresponding endian path.

> The original code probably used unsafe for a reasonable historical reason: it is a direct native-endian
  reinterpretation and can avoid conversion overhead or allocations in hot scalar paths, especially before
  the scratch-buffer approach. With the scratch buffer in place, the measurable difference is negligible in
  reads, writes stay allocation-free with encoding/binary, and the code becomes clearer, portable, and less
  dependent on unsafe aliasing/alignment assumptions.

HashimTheArab:
No issues on prod (140 players)

PR #431, opened 2026-05-17
Avoid protocol string marshal allocations (0 alloc, 3.27x faster)

HashimTheArab:
```go
func BenchmarkWriterStringBytesBuffer(b *testing.B) {
	var buf bytes.Buffer
	var writer Writer
	writer.Reset(&buf, 0)
	s := strings.Repeat("gophertunnel", 16)

	b.ReportAllocs()
	b.SetBytes(int64(len(s)))
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		buf.Reset()
		writer.String(&s)
	}
}
```

BenchmarkWriterStringBytesBuffer

Before: 41.42 ns/op, 192 B/op, 1 alloc/op
After:  12.68 ns/op,   0 B/op, 0 allocs/op
Delta: -69.40% sec/op, -100% B/op, -100% allocs/op
Speedup: 41.42 / 12.68 = ~3.27x faster

### DIFF
--- a/minecraft/protocol/writer.go
+++ b/minecraft/protocol/writer.go
@@ -76,10 +76,12 @@ func (w *Writer) String(x *string) {
 	w.writeString(*x)
 }
 
+// stringWriter is implemented by writers that support writing strings directly.
 type stringWriter interface {
 	WriteString(string) (int, error)
 }
 
+// writeString uses WriteString when available to avoid converting s to []byte.
 func (w *Writer) writeString(s string) {
 	if sw, ok := w.w.(stringWriter); ok {
 		_, _ = sw.WriteString(s)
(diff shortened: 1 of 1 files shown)


minecraft/protocol/writer.go
```diff
@@ -344,7 +367,8 @@ func (w *Writer) ItemInstance(i *ItemInstance) {
 		internal.BufferPool.Put(buf)
 	}()
 
-	bufWriter := NewWriter(buf, w.shieldID)
+	var bufWriter Writer
+	bufWriter.Reset(buf, w.shieldID)
```
TwistedAsylumMC: For what reason has this changed? Surely it would be better to keep the NewWriter call?

TwistedAsylumMC: Same for everything else below, why create a non-pointer value to then reference it as a pointer later on?

TwistedAsylumMC (changes requested):

PR #434, opened 2026-05-19
Implement connecting to worlds (take 2)

lactyy2:
This PR adds functionality to support connecting to friend worlds, along with other protocol changes required to successfully negotiate with vanilla hosts.
The `minecraft/p2p` package includes a client for listing and connecting to peer-to-peer worlds hosted by players.

Hosting a world (acting as a listener) is not yet supported. It requires syncing the server status (player count and MOTD) with the world description published to Xbox Live, and it is not yet clear how this should be done.

This PR also migrates to use `go-xsapi/v2` and a new version of `go-playfab` to authenticate with Minecraft network services.


lactyy2:
This PR is ready for review. The workflow requires approval because I updated the Go version.

minecraft/nethernet.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+	"net"
+
+	"github.com/df-mc/go-nethernet"
+)
+
```
HashimTheArab, line 13: It's helpful to have a configurable Dialer here like the previous PR, you can probably copy and paste this file:
https://github.com/HashimTheArab/gophertunnel/blob/ec3bd305af806681e312ff76a5fbc1d2da3cc1a6/minecraft/nethernet.go

minecraft/service/signaling/message.go
```diff
@@ -0,0 +1,129 @@
...
+		b.WriteString(strconv.Quote(err.Message))
+	}
+	return b.String()
+}
+
+const (
+	// ErrorCodePlayerNotFound indicates that the remote network ID specified
+	// in [Message.To] is not found. It is currently unused, and ErrorCodeDeliveryFailure
+	// is instead used for this purpose.
+	ErrorCodePlayerNotFound = iota + 1
```
HashimTheArab, line 66: is this correct? in the old code it was 1 https://github.com/HashimTheArab/gophertunnel/blob/ec3bd305af806681e312ff76a5fbc1d2da3cc1a6/minecraft/service/signaling/message.go#L58

minecraft/service/signaling/dial.go
```diff
@@ -0,0 +1,138 @@
...
+	}{Alias: (*Alias)(e)}
+	if err := json.Unmarshal(b, &data); err != nil {
+		return err
+	}
+	e.ServiceURI, err = url.Parse(data.ServiceURI)
+	if err != nil {
+		return fmt.Errorf("parse service URI: %w", err)
+	}
+	return nil
+}
```
HashimTheArab: Maybe this makes more sense in environment.go like it was before, since its also imported by jsonrpc

minecraft/nethernet.go
lactyy2: added in [`12d3c3d`](https://github.com/Sandertv/gophertunnel/pull/434/commits/12d3c3de77b87b81d29b144ed151aaa13b11a45b)

minecraft/service/signaling/dial.go
lactyy2: resolved in [`7c69b5a`](https://github.com/Sandertv/gophertunnel/pull/434/commits/7c69b5a732de21e57e8fabba10dc8e21198a1b24)

minecraft/nethernet.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"errors"
+	"net"
+
+	"github.com/df-mc/go-nethernet"
+)
+
+// NetherNet is an implementation of NetherNet network. Unlike RakNet, it needs to be registered manually with a Signaling.
```
HashimTheArab:
```suggestion
// NetherNet is an implementation of a NetherNet network using a Signaling backend.
```

minecraft/service/signaling/message.go
lactyy2: i debugged this and confirmed that it is correct

```
error sending signal: service/signaling: code 1: "Player not found."
```

minecraft/service/internal/result.go
```diff
@@ -50,3 +52,36 @@ func Err(resp *http.Response) error {
...
+// on the [ServiceEnvironment.ServiceURI].
+type ServiceEnvironment struct {
+	ServiceURI *url.URL `json:"serviceUri"`
+}
+
+// UnmarshalJSON implements [json.Unmarshaler.UnmarshalJSON].
+// Since [url.URL] does not implement [json.Unmarshaler] or [encoding.TextUnmarshaler],
+// we decode URL fields as strings first, then manually parse them into URLs.
+// See: https://github.com/golang/go/issues/52638
+func (e *ServiceEnvironment) UnmarshalJSON(b []byte) error {
```
HashimTheArab, line 66: this method can be simplified
```go
 func (e *ServiceEnvironment) UnmarshalJSON(b []byte) error {
      var data struct {
              ServiceURI string `json:"serviceUri"`
      }
      if err := json.Unmarshal(b, &data); err != nil {
              return err
      }
      if data.ServiceURI == "" {
              return errors.New("service/internal: ServiceEnvironment.ServiceURI is empty")
      }
      u, err := url.Parse(data.ServiceURI)
      if err != nil {
              return fmt.Errorf("parse ServiceURI: %w", err)
      }
      e.ServiceURI = u
      return nil
}

```
  - ServiceEnvironment only has one field: ServiceURI.
  - The anonymous Alias pointer only exists to let JSON populate other fields on ServiceEnvironment, but there are no other fields.

  - Issuer string is unused. Since json.Unmarshal ignores unknown JSON fields by default, removing Issuer does not change handling of JSON containing "issuer".

  - The current ServiceURI string field shadows the embedded Alias.ServiceURI, so even today ServiceURI is decoded as a string and then manually parsed into *url.URL.

HashimTheArab: Nevermind I saw that this pattern is consistent with all of the other UnmarshalJSON implementations

minecraft/conn.go
```diff
@@ -893,9 +891,18 @@ func (conn *Conn) handleServerToClientHandshake(pk *packet.ServerToClientHandsha
...
+		return fmt.Errorf("convert private key to ECDH: %w", err)
+	}
+	publicKey, err := pub.ECDH()
+	if err != nil {
+		return fmt.Errorf("convert public key to ECDH: %w", err)
+	}
+	sharedSecret, err := privateKey.ECDH(publicKey)
+	if err != nil {
+		return fmt.Errorf("compute shared secret: %w", err)
+	}
```
HashimTheArab: Consider a helper for this since its duplicated below aswell
```go
func (conn *Conn) encryptionKey(salt []byte, pub *ecdsa.PublicKey) ([32]byte, error) {
      privateKey, err := conn.privateKey.ECDH()
      if err != nil {
              return [32]byte{}, fmt.Errorf("convert private key to ECDH: %w", err)
      }
      publicKey, err := pub.ECDH()
      if err != nil {
              return [32]byte{}, fmt.Errorf("convert public key to ECDH: %w", err)
      }
      sharedSecret, err := privateKey.ECDH(publicKey)
      if err != nil {
              return [32]byte{}, fmt.Errorf("compute shared secret: %w", err)
      }
      return sha256.Sum256(append(salt, sharedSecret...)), nil
}
```

minecraft/p2p/client.go
```diff
@@ -0,0 +1,216 @@
...
+				s.log.Debug("received nonce from host", slog.String("xuid", xuid), slog.String("nonce", nonce))
+				s.nonce = nonce
+				close(s.nonceReceived)
+			}
+		}
+	}
+	return nil
+}
+
+// Connection returns the first entry on [World.SupportedConnections].
```
HashimTheArab: the "first entry" is outdated now

minecraft/p2p/client.go
```diff
@@ -0,0 +1,216 @@
...
+// [github.com/sandertv/gophertunnel/minecraft/protocol/login.ClientData.Nonce]
+// field.
+func (s *Session) Nonce() string {
+	s.worldMu.RLock()
+	defer s.worldMu.RUnlock()
+
+	return s.nonce
+}
+
+// Close leaves the session with a 15 seconds timeout.
```
HashimTheArab, line 234: this comment is outdated

minecraft/p2p/client.go
```diff
@@ -0,0 +1,216 @@
...
+		// If the host has not yet generated or published a nonce for the caller, check if
+		// one has been added to the Nonces field since the last update.
+		xuid := s.client.client.UserInfo().XUID
+		if nonces := world.Nonces; nonces != nil {
+			nonce, ok := nonces[xuid]
+			if ok {
+				if nonce == "" {
+					return errors.New("host published empty nonce for caller")
+				}
+				s.log.Debug("received nonce from host", slog.String("xuid", xuid), slog.String("nonce", nonce))
```
HashimTheArab, line 171: not sure if nonce should be in debug logs since its auth data but either way is fine

minecraft/p2p/client.go
```diff
@@ -0,0 +1,216 @@
...
+	return &Client{
+		client: client,
+		conf:   conf,
+	}
+}
+
+// NewClient returns a new Client using the underlying [xsapi.Client].
+func NewClient(client *xsapi.Client) *Client {
+	var c ClientConfig
+	return c.New(client)
```
HashimTheArab: could be simplified to `return ClientConfig{}.New(client)` but up to you

minecraft/p2p/client.go
```diff
@@ -0,0 +1,216 @@
...
+
+// handler handles events that might occur in the Session.
+type handler struct {
+	*Session
+}
+
+// HandleSessionChange handles a session change notified by the RTA service.
+func (h *handler) HandleSessionChange(session *mpsd.Session) {
+	if err := h.updateWorldData(session.Properties().Custom); err != nil {
+		h.log.Error("error updating world data", slog.Any("error", err))
```
HashimTheArab, line 252: If the host sends an empty nonce for us, Join currently waits until timeout instead of returning the error, because HandleSessionChange only logs updateWorldData errors. Should we return that error from Join?

minecraft/p2p/client.go
lactyy2: that nonce is visible to all players even who doesn't have proper access to that multiplayer session because Xbox Live also exposes custom properties from multiplayer handles

HashimTheArab: Oh nevermind then I misunderstood it

HashimTheArab (approved):

lactyy2:
With the help of @HashimTheArab, this PR is now ready.

HashimTheArab (approved):

PR #436, opened 2026-05-21
Guard packet reader allocations from impossible lengths

HashimTheArab:
This fixes an issue I ran into where an incorrect BiomeDefinitionList packet schema caused a malformed length prefix to be decoded as a huge slice length, leading to an attempted ~1 TB allocation before the packet failed to decode.

The reader now validates string, byte-slice, and slice lengths against the remaining packet payload before allocating.


cqdetdev:
Could you please document this 1 terabyte of allocation please.

HashimTheArab:
image
i sent it somewhere before heres 192gb

PR #437, opened 2026-05-27
Make snappy 3-10x faster and 0 allocations, 28x faster than flate

HashimTheArab:
- Switches Snappy compression from `snappy.Encode` to `s2.EncodeSnappy`, keeping Snappy-compatible output while making the hot encode path much faster.
- Adds an append-based Snappy path so compressed data is written into the encoder output buffer directly, avoiding the extra compressed slice/copy.
- Eliminates Snappy encode allocations for compressed batches that stay within the `1 MiB` pooled encoder buffer cap.

In benchmarks, the new Snappy path reaches ~5.98-35.89 Gbps effective encode throughput, is 2.5x-9.6x faster than the previous Snappy path, and up to ~28x faster than Flate depending on batch size

image

Benchmark code
```go
package packet

import (
      "io"
      "testing"
)

func BenchmarkEncoderCompression(b *testing.B) {
      cases := []struct {
              name       string
              packets    int
              packetSize int
      }{
              {name: "1x1KiB", packets: 1, packetSize: 1 << 10},
              {name: "16x1KiB", packets: 16, packetSize: 1 << 10},
              {name: "1x128KiB", packets: 1, packetSize: 128 << 10},
              {name: "4x128KiB", packets: 4, packetSize: 128 << 10},
              {name: "8x128KiB", packets: 8, packetSize: 128 << 10},
      }

      compressions := []struct {
              name        string
              compression Compression
      }{
              {name: "Snappy", compression: SnappyCompression},
              {name: "Flate", compression: FlateCompression},
      }

      for _, c := range cases {
              packets, total := benchmarkPackets(c.packets, c.packetSize)

              for _, compression := range compressions {
                      b.Run(c.name+"/"+compression.name, func(b *testing.B) {
                              encoder := NewEncoder(io.Discard)
                              encoder.EnableCompression(compression.compression, 0)

                              b.ReportAllocs()
                              b.SetBytes(int64(total))
                              b.ResetTimer()

                              for i := 0; i < b.N; i++ {
                                      if err := encoder.Encode(packets); err != nil {
                                              b.Fatal(err)
                                      }
                              }
                      })
              }
      }
}

func benchmarkPackets(n, size int) ([][]byte, int) {
      packets := make([][]byte, n)
      total := 0

      for i := range packets {
              p := make([]byte, size)
              for j := range p {
                      p[j] = byte((j*31 + i*17) & 0xff)
              }
              packets[i] = p
              total += len(p)
      }

      return packets, total
}
```

### DIFF
--- a/minecraft/protocol/packet/compression.go
+++ b/minecraft/protocol/packet/compression.go
@@ -5,10 +5,11 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"slices"
 	"sync"
 
 	"github.com/klauspost/compress/flate"
-	"github.com/klauspost/compress/snappy"
+	"github.com/klauspost/compress/s2"
 	"github.com/sandertv/gophertunnel/minecraft/internal"
 )
 
@@ -20,6 +21,10 @@ type Compression interface {
 	Compress(decompressed []byte) ([]byte, error)
 	// Decompress decompresses the given data and returns the decompressed data.
 	Decompress(compressed []byte, limit int) ([]byte, error)
+}
+
+type appendCompression interface {
+	CompressAppend(dst, decompressed []byte) ([]byte, error)
 }
 
 var (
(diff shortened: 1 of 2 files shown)


HashimTheArab:
1 gb seems to be a good default for the pool size cap so this PR is ready for review

image

HashimTheArab:
hi can we merge pls thx

PR #438, opened 2026-06-02
gophertunnel: Updated for 1.26.30.

didntpot:


minecraft/conn.go
```diff
@@ -1169,7 +1169,7 @@ func (conn *Conn) handleResourcePackDataInfo(pk *packet.ResourcePackDataInfo) er
 		for i := uint32(0); i < chunkCount; i++ {
 			_ = conn.WritePacket(&packet.ResourcePackChunkRequest{
 				UUID:       idCopy,
-				ChunkIndex: i,
+				ChunkIndex: int32(i),
```
TwistedAsylumMC: Only nitpick, I would make chunkCount an int32 so you dont need this cast

PR #444, opened 2026-06-20
Minimal changes to support using NetherNet as a transport

lactyy2:


minecraft/protocol/login/data.go
```diff
@@ -345,7 +346,22 @@ func (data ClientData) Validate() error {
...
+		if u.Host == "" || u.Port() == "" || u.Port() != data.ServerAddress[ind+1:] ||
+			(u.Scheme != "https" && u.Scheme != "http") {
+			return fmt.Errorf("ServerAddress is invalid: %v", data.ServerAddress)
+		}
+		if _, err := net.ResolveUDPAddr("udp", u.Host); err != nil {
+			return fmt.Errorf("ServerAddress must be resolveable as a UDP address, but got %v", data.ServerAddress)
+		}
+	} else if _, err := net.ResolveUDPAddr("udp", data.ServerAddress); err != nil {
 		return fmt.Errorf("ServerAddress must be resolveable as a UDP address, but got %v", data.ServerAddress)
 	}
```
TwistedAsylumMC: Do these need to be duplicated? You're also not passing in u.Host to the error

TwistedAsylumMC (changes requested):
Just one minor comment

minecraft/protocol/login/data.go
lactyy2: I think validating the URL should be enough. Removed that check in [`42b3cf8`](https://github.com/Sandertv/gophertunnel/pull/444/commits/42b3cf85341e98a5583b6e1da7e14287a9df1895).

HashimTheArab:
If we can get https://github.com/Sandertv/gophertunnel/pull/445 merged first it would be helpful to add this in minecraft/nethernet.go
```
// Ensure the connection returned by NetherNet.DialContext has the optional
// packet methods used by Encoder and Decoder, even though DialContext returns it
// as a net.Conn.
var _ packet.TransportCapabilities = (*nethernet.Conn)(nil)
```

lactyy2:
Added an example test demonstrating a NetherNet<->NetherNet MITM proxy in [minecraft/example_nethernet_test.go](https://github.com/lactyy2/gophertunnel/blob/feature/nethernet/minecraft/example_nethernet_test.go).

PR #446, opened 2026-06-22
Pace resource pack chunks for clients after 1.26.30

HashimTheArab:
Paces `ResourcePackChunkData` writes by 200ms so clients after 1.26.30 are less likely to fail resource pack downloads when chunks are sent in a burst.

This follows the same compatibility fix Geyser applied in GeyserMC/Geyser@ee8e13727a277e81ab9bf4e1424b3ae234dc9f71.

### DIFF
--- a/minecraft/conn.go
+++ b/minecraft/conn.go
@@ -1010,8 +1010,14 @@ func (conn *Conn) hasPack(uuid string, version string, hasBehaviours bool) bool 
 	return false
 }
 
-// packChunkSize is the size of a single chunk of data from a resource pack: 512 kB or 0.5 MB
-const packChunkSize = 1024 * 128
+const (
+	// packChunkSize is the size of a single chunk of data from a resource pack: 128 KiB.
+	packChunkSize = 1024 * 128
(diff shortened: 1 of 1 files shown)


smell-of-curry:
Production data point in favor of this: we run a dragonfly hub plus gophertunnel-based downstream proxies (both on v1.57.1) serving a ~52 MB pack, and a subset of players — disproportionately consoles and high-latency clients — stall at 0/N chunks on the download screen until they give up. Local pack install sidesteps it, which matches the burst-overload theory rather than a pack problem.

Worth noting every other major implementation already ships an equivalent of this after 1.21.30 clients started failing on bursts: Geyser (256 KiB chunks + 200 ms spacing, the commit referenced here), PocketMine-MP (256 KiB + ACK-throttled, one chunk in flight), PowerNukkitX and WaterdogPE similarly. gophertunnel is the outlier still writing all chunks back-to-back as fast as the client requests them, so anything built on it (dragonfly included) inherits the stalls. Would love to see this merged.

PR #463, opened 2026-07-10
Fix NetherNet example client construction

HashimTheArab:
Use the current go-nethernet client API so go vet passes.


HashimTheArab:
wrong repo wrong branch wrong changes gpt 5.6 luna sucks

PR #480, opened 2026-08-04
minecraft/protocol: Support 1.26.40 (protocol 2168)

smell-of-curry:
Adds support for Minecraft Bedrock 1.26.40, protocol 2168.

## How this was verified

Everything here was checked against a real 1.26.40.8 dedicated server rather than only against the changelog, because several documented changes did not match what the server actually sends. A probe client connected, played through chat, commands, block breaking and placing, item use, inventory requests, scoreboards, sounds, sub chunk requests and map items, and captured every packet in both directions. Two things were then required of each capture: the connection had to complete with zero decode errors and zero packet violations, and every captured payload had to decode and re-encode to the exact bytes the server sent.

The final run decoded 65 distinct client bound packet types with no decode errors and no violations, and 88 packet types round trip byte for byte.

## Where the server disagrees with the documentation

Two of these are worth calling out, since following the documentation alone produces a client the server rejects.

Recipe ingredients still name their descriptor type with a string. The changelog describes a constant byte, but a `CraftingData` payload from this server contains the literal string `item_tag` 245 times. Item descriptors inside an `ItemStackRequest` are a genuinely different encoding, selecting the descriptor by index and counting with a little endian `uint16`, so both forms are implemented and both are exercised against the server.

Stack request actions select their variant by its index in the variant list. `PlaceInContainer` and `TakeOutContainer` were removed from that list but the types after them kept the values they had while they were there, so from `LabTableCombine` onwards the index runs two behind the type value that immediately follows it on the wire. Writing the value in both places, as the previous code did, made the server pick the wrong variant and reject the request with `wrong const value for member "Action type"`.

## Changes

`CraftingData` is now twelve segmented recipe vectors rather than one tagged vector, so the `Recipe` interface and its lookup functions are gone, and a recipe's unlock requirement is optional. The deprecated craft result action carries its own item form, added alongside the descriptor form as `StackReqItem` and `StackReqItemDescriptorCount`.

`SetScore` entries are a tagged variant, and the scoreboard identity entry's player ID is optional. `SubChunk` positions are plain `int32`s, matching `SubChunkRequest`. Map colours are fixed four byte values, and a tracked object's unique ID and block position are each preceded by a presence byte. `MoveActorDelta` replaces its packed flag header with optional fields, and `MovePlayer` carries optional teleport data. `AnvilDamage` drops its damage amount. `StructureBlockUpdate` reads the redstone save mode as a byte and its name as a redactable string. `PresenceInfo` keeps only an optional rich presence ID. `Transfer` gained an optional gatherings configuration, `ServerBoundDiagnostics` a system category array, and `TextShape` a line gap height. `ClientboundUpdateSoundData` is new, holding seven optional sound updates.

Also adds the OreUI client memory category, the mount, dismount and straw bed break sound events, and the entity flag for not being pickable from inside.

## Unrelated observations

Two things turned up during testing that are not 1.26.40 changes and are left alone here. `Disconnect` and `ContainerClose` are both absent from the client pool although a vanilla client sends both, so a proxy cannot decode them. And `CommandRequest`'s version field is a string in this library where the server reads an `int32`, which makes a request with an empty version fail as exceeding the maximum command length.


minecraft/protocol/packet/move_player.go
```diff
@@ -46,11 +46,7 @@ type MovePlayer struct {
 	// RiddenEntityRuntimeID is the runtime ID of the entity that the player might currently be riding. If not
 	// riding, this should be left 0.
 	RiddenEntityRuntimeID uint64
-	// TeleportCause is written only if Mode is MoveModeTeleport. It specifies the cause of the teleportation,
-	// which is one of the constants above.
-	TeleportCause int32
-	// TeleportSourceEntityType is the entity type that caused the teleportation, for example an ender pearl.
-	TeleportSourceEntityType int32
+	TeleportData          protocol.Optional[protocol.TeleportData]
```
TwistedAsylumMC, line 50: Needs doc

minecraft/protocol/packet/player_auth_input.go
```diff
@@ -76,22 +76,22 @@ const (
...
+	PlayModeTeaser
+	PlayModeScreen
+	_
+	_
+	_
+	_
+	PlayModeExitLevel
+	_
+	PlayModeNumModes
 )
```
TwistedAsylumMC, line 98: Whats with the formatting

minecraft/protocol/packet/player_auth_input.go
```diff
@@ -168,32 +168,29 @@ func (pk *PlayerAuthInput) Marshal(io protocol.IO) {
...
+	io.Bool(&t)
+	protocol.OptionalMarshaler(io, &pk.ItemStackRequest)
+	io.Bool(&t)
+	protocol.OptionalFunc(io, &pk.BlockActions, func(x *[]protocol.PlayerBlockAction) {
+		protocol.Slice(io, x)
+	})
+	io.Bool(&t)
+	protocol.OptionalFunc(io, &pk.VehicleRotation, io.Vec2)
+	io.Bool(&t)
+	protocol.OptionalFunc(io, &pk.ClientPredictedVehicle, io.Varint64)
```
TwistedAsylumMC: This will not decode properly if a server sends false for the first bool and nothing else

minecraft/protocol/packet/player_location.go
```diff
@@ -30,8 +30,13 @@ func (*PlayerLocation) ID() uint32 {
 }
 
 func (pk *PlayerLocation) Marshal(io protocol.IO) {
-	io.Int32(&pk.Type)
 	io.Varint64(&pk.EntityUniqueID)
+	t := uint32(pk.Type)
+	io.Varuint32(&t)
+	pk.Type = int32(t)
+	// Cereal leftover field; always 0 on the wire.
```
TwistedAsylumMC: Don't think this is actually true, just an issue with the docs. The second type is the same as before

minecraft/protocol/packet/resource_pack_client_response.go
```diff
@@ -28,6 +28,34 @@ func (*ResourcePackClientResponse) ID() uint32 {
...
-	protocol.FuncSliceUint16Length(io, &pk.PacksToDownload, io.String)
+	// As of 1.26.400 the response is a varuint32 that is zero-indexed: the value on the wire is the
+	// response constant minus one. Adjust in both directions so pk.Response keeps using the 1-based
+	// constants above.
+	response := uint32(pk.Response) - 1
+	io.Varuint32(&response)
+	pk.Response = byte(response + 1)
+
+	// A varuint32 length-prefixed string follows the response. Its contents are ignored by the client,
+	// so we write the readable name of the response for parity and discard whatever is read.
```
TwistedAsylumMC: Please remove slop comments like these

TwistedAsylumMC: Also why no just change the enums here?

minecraft/protocol/map.go
```diff
@@ -52,13 +52,16 @@ type MapTrackedObject struct {
...
+	if entity {
 		r.Varint64(&x.EntityUniqueID)
-	case MapObjectTypeBlock:
+	}
+	block := x.Type == MapObjectTypeBlock
+	r.Bool(&block)
+	if block {
 		r.BlockPos(&x.BlockPosition)
-	default:
-		r.UnknownEnumOption(x.Type, "map tracked object type")
```
TwistedAsylumMC: Don't hack in stuff based on existing types. Feel free to change for this to be better

minecraft/protocol/player.go
```diff
@@ -84,7 +92,20 @@ type PlayerListEntry struct {
 
 // Marshal encodes/decodes a PlayerListEntry.
 func (x *PlayerListEntry) Marshal(r IO) {
+	// The action is sent twice: first as the index of the entry variant (which is
+	// the inverse of ActionType, as an added entry is variant 1) and then as the
+	// action itself.
```
TwistedAsylumMC: slop comment

minecraft/protocol/player.go
```diff
@@ -84,7 +92,20 @@ type PlayerListEntry struct {
...
+	variant := uint32(0)
+	if x.ActionType == PlayerListActionAdd {
+		variant = 1
+	}
+	r.Varuint32(&variant)
```
TwistedAsylumMC: probably make this a separate function to keep it clean

minecraft/protocol/reader.go
```diff
@@ -268,36 +268,22 @@ func (r *Reader) UUID(x *uuid.UUID) {
 // PlayerInventoryAction reads a PlayerInventoryAction.
 func (r *Reader) PlayerInventoryAction(x *UseItemTransactionData) {
 	r.Varint32(&x.LegacyRequestID)
-	if x.LegacyRequestID < -1 && (x.LegacyRequestID&1) == 0 {
+	var legacy bool
+	r.Bool(&legacy)
+	if legacy && x.LegacyRequestID < -1 && (x.LegacyRequestID&1) == 0 {
```
TwistedAsylumMC: remove everything other than bool

TwistedAsylumMC: actually, make the legacy slice optional

minecraft/protocol/reader.go
```diff
@@ -268,36 +268,22 @@ func (r *Reader) UUID(x *uuid.UUID) {
...
-		x.WindowID = int8(windowID)
-	case InventoryActionSourceWorld:
-		r.Varuint32(&x.SourceFlags)
+	// The actions are wrapped in a struct that is itself optional, so their
+	// presence is stated by two booleans rather than one.
+	var outer, inner bool
+	r.Bool(&outer)
+	r.Bool(&inner)
+	if outer && inner {
+		Slice(r, &x.Actions)
```
TwistedAsylumMC: Maybe we should add a temporary double optional function, it will only be short lived since these are all fixed in U5

minecraft/protocol/reader.go
```diff
@@ -268,36 +268,22 @@ func (r *Reader) UUID(x *uuid.UUID) {
...
 	}
-	r.Varuint32(&x.InventorySlot)
-	r.ItemInstance(&x.OldItem)
-	r.ItemInstance(&x.NewItem)
+	// Everything from the action type onwards is identical to the transaction
+	// data carried by an InventoryTransaction packet.
```
TwistedAsylumMC: Slop

minecraft/protocol/reader.go
```diff
@@ -308,6 +294,9 @@ func (r *Reader) GameRule(x *GameRule) {
 	r.Varuint32(&t)
 
 	switch t {
+	case 0:
+		// Null variant: no value payload follows.
```
TwistedAsylumMC: slop

minecraft/protocol/reader.go
```diff
@@ -360,6 +324,9 @@ func (r *Reader) EntityMetadata(x *EntityMetadata) {
 		var key, dataType uint32
 		r.Varuint32(&key)
 		r.Varuint32(&dataType)
+		// The type of the value is sent a second time as a byte.
```
TwistedAsylumMC: slop

minecraft/protocol/reader.go
```diff
@@ -405,24 +372,28 @@ func (r *Reader) EntityMetadata(x *EntityMetadata) {
 
 // ItemDescriptorCount reads an ItemDescriptorCount i from the underlying buffer.
 func (r *Reader) ItemDescriptorCount(i *ItemDescriptorCount) {
-	var id uint8
-	r.Uint8(&id)
+	var present uint32
+	r.Varuint32(&present)
```
TwistedAsylumMC: why is "present" a number? could be a bool if it's only ever 0 or 1

minecraft/protocol/reader.go
```diff
@@ -515,102 +424,134 @@ func (r *Reader) ItemInstanceNew(i *ItemInstance) {
...
+	*blockingTick = 0
+	if blocking {
+		bufReader.Int64(blockingTick)
 	}
 }
 
 // StackRequestAction reads a StackRequestAction from the reader.
 func (r *Reader) StackRequestAction(x *StackRequestAction) {
+	// The variant is selected by its index in the variant list, and the type is
+	// then repeated as the variant's own first field.
```
TwistedAsylumMC, line 547: slop

minecraft/protocol/scoreboard.go
```diff
@@ -1,58 +1,91 @@
...
+	ScoreboardActionNameChangeFakePlayer = "changefakeplayer"
+)
+
+// scoreboardActionNames maps the actions above to their name on the wire, in action order.
+var scoreboardActionNames = [...]string{
+	ScoreboardActionNameRemove,
+	ScoreboardActionNameChangePlayer,
+	ScoreboardActionNameChangeEntity,
+	ScoreboardActionNameChangeFakePlayer,
+}
```
TwistedAsylumMC: Not consistent with other ways of doing this

minecraft/protocol/shape.go
```diff
@@ -112,6 +112,8 @@ type TextShape struct {
 	UseRotation bool
 	// BackgroundColour is the RGBA colour to use for the text background. This is a translucent black colour by default.
 	BackgroundColour Optional[color.RGBA]
+	// LineGapHeight is the gap between lines used when rendering multiline text.
+	LineGapHeight float32
```
TwistedAsylumMC: This is not in the update

minecraft/protocol/skin.go
```diff
@@ -72,6 +78,8 @@ type Skin struct {
 	// OverrideAppearance specifies if the skin should override the player's skin that is equipped client-side.
 	// When false, the client will reject the skin and continue to use the skin that the player has equipped.
 	OverrideAppearance bool
+	TrustedSkinFlag    string
+	ProfileHash        string
```
TwistedAsylumMC: Docs

minecraft/protocol/skin.go
```diff
@@ -55,12 +63,10 @@ type Skin struct {
 	// FullID is an ID that represents the skin in full. The actual functionality is unknown: The client
 	// does not seem to send a value for this.
 	FullID string
-	// SkinColour is a hex representation (including #) of the base colour of the skin. An example of the
-	// colour sent here is '#b37b62'.
-	SkinColour string
-	// ArmSize is the size of the arms of the player's model. This is either 'wide' (generally for male skins)
-	// or 'slim' (generally for female skins).
-	ArmSize string
+	// SkinColour is a hex representation (including #) of the base colour of the skin.
```
TwistedAsylumMC: Cant have a #

minecraft/protocol/skin.go
```diff
@@ -220,6 +249,6 @@ type PersonaPieceTintColour struct {
 
 // Marshal encodes/decodes a PersonaPieceTintColour.
 func (x *PersonaPieceTintColour) Marshal(r IO) {
-	r.String(&x.PieceType)
+	r.Uint32(&x.PieceType)
 	FuncSliceUint32Length(r, &x.Colours, r.String)
```
TwistedAsylumMC: Colours has no length prefix, just always writes 4 values

minecraft/protocol/sub_chunk.go
```diff
@@ -47,35 +47,42 @@ type SubChunkEntry struct {
...
 
 // SubChunkEntryNoCache encodes/decodes a SubChunkEntry assuming the blob cache is not enabled.
 func SubChunkEntryNoCache(r IO, x *SubChunkEntry) {
+	subChunkEntry(r, x, false)
+}
+
+// subChunkEntry encodes/decodes a SubChunkEntry. Every field that used to be conditional now carries its own
+// presence byte, so the presence read from the stream is what decides whether a field follows, while encoding
+// derives it from the result and height map types it belongs to.
+func subChunkEntry(r IO, x *SubChunkEntry, cache bool) {
```
TwistedAsylumMC: Similar to before, rework the types rather than making the marshaling messy. Probably dont need this helper function either if the cache field is at the end

minecraft/protocol/writer.go
```diff
@@ -193,35 +193,20 @@ func (w *Writer) UUID(x *uuid.UUID) {
...
-	}
-	w.Varuint32(&x.InventorySlot)
-	w.ItemInstance(&x.OldItem)
-	w.ItemInstance(&x.NewItem)
+	// The actions are wrapped in a struct that is itself optional, so their
+	// presence is stated by two booleans rather than one.
+	present := true
+	w.Bool(&present)
+	w.Bool(&present)
+	Slice(w, &x.Actions)
```
TwistedAsylumMC: Double optional thing again

TwistedAsylumMC (changes requested):
Went through most of it thoroughly, just a few main comments. Biggest thing is just the comments inside of encoding methods, 99% of the time not necessary. I didn't comment on all of them, but please remove them unless absolutely necessary (not likely)

minecraft/protocol/packet/move_player.go
smell-of-curry: Added field docs for `TeleportData` and documented `protocol.TeleportData`.

minecraft/protocol/packet/player_auth_input.go
smell-of-curry: Fixed: added temporary `DoubleOptionalFunc` and wired all five nested optionals through it so an outer `false` no longer reads an inner bool. Also switched `InputData` to `InputFlags`, named the PlayMode gaps, and added `InputFlagInternalUpdate`.

minecraft/protocol/packet/player_auth_input.go
smell-of-curry: Fixed: added temporary `DoubleOptionalFunc` and wired all five nested optionals through it so an outer `false` no longer reads an inner bool. Also switched `InputData` to `InputFlags`, named the PlayMode gaps, and added `InputFlagInternalUpdate`.

minecraft/protocol/packet/player_location.go
smell-of-curry: Kept the reserved `Varint32` — Cloudburst v2168 still writes/reads a zero here after the type. Dropped the misleading leftover/docs framing; the wire field is real even if the Mojang docs look like a duplicated type.

minecraft/protocol/packet/resource_pack_client_response.go
smell-of-curry: Done: enums are now 0-based iota, Marshal writes them directly, and the wire-format comments are gone.

minecraft/protocol/map.go
smell-of-curry: Reworked `MapTrackedObject` to independent `Optional` fields instead of deriving presence from `Type`.

minecraft/protocol/player.go
smell-of-curry: Removed the comment and extracted `playerListAction`.

minecraft/protocol/player.go
smell-of-curry: Removed the comment and extracted `playerListAction`.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/reader.go
smell-of-curry: Made `LegacySetItemSlots` a plain `Optional`, Actions go through `DoubleOptionalFunc`, item-descriptor presence is a bool, and the Marshal-body slop comments are gone.

minecraft/protocol/scoreboard.go
smell-of-curry: Restyled to match the library's other variant patterns (`IdentityType` + optional objective on remove).

minecraft/protocol/shape.go
smell-of-curry: Removed `LineGapHeight` — not part of this update.

minecraft/protocol/skin.go
smell-of-curry: Docs fixed (no `#` on `SkinColour`), `ProfileHash` documented, and tint `Colours` are a fixed 4× `BEARGB` with no length prefix.

minecraft/protocol/skin.go
smell-of-curry: Docs fixed (no `#` on `SkinColour`), `ProfileHash` documented, and tint `Colours` are a fixed 4× `BEARGB` with no length prefix.

minecraft/protocol/skin.go
smell-of-curry: Docs fixed (no `#` on `SkinColour`), `ProfileHash` documented, and tint `Colours` are a fixed 4× `BEARGB` with no length prefix.

minecraft/protocol/sub_chunk.go
smell-of-curry: Reworked `SubChunkEntry` to Optionals with a single `Marshal`; dropped the NoCache helper.

minecraft/protocol/writer.go
smell-of-curry: Same double-optional / legacy-optional cleanup as the reader.

smell-of-curry:
Addressed all of @TwistedAsylumMC's review comments in dd3d6fd.

Highlights:
- Temporary `DoubleOptionalFunc` for PlayerAuthInput nested optionals (outer false no longer mis-decodes)
- `LegacySetItemSlots` / Actions cleaned up the same way
- Pack response enums 0-based; MapTrackedObject / SubChunkEntry reworked to Optionals
- Skin tint colours fixed-length 4× BEARGB; LineGapHeight removed
- Marshal-body wire-format comments stripped

Also absorbed the remaining gaps vs #481 (InputFlags, particle constants, player_update_entity_overrides cereal tags, PlayerPermissions byte, GatheringJoinInfo optionals) while keeping the live-verified CraftingData string descriptors, stack-request action variant remap, and sound Fade Duration-before-TargetVolume order.

Re-probed against 1.26.40.8 BDS: 0 decode errors, 0 violations, 88 packet types round-trip exactly.

TwistedAsylumMC:
Closed in favour of https://github.com/Sandertv/gophertunnel/pull/481. Thank you for the PR though!

PR #481, opened 2026-08-04
minecraft: update protocol to 1.26.40

HashimTheArab:
## Summary

- Update the supported Minecraft version to 1.26.40 / protocol 2168.
- Implement the v2168 Cereal discriminators and optional wrappers across entity metadata, inventory, movement, player, score, map, skin, sound, chunk, and resource-pack payloads.
- Migrate crafting data to the v2168 typed recipe vectors and update item/descriptor wire formats.
- Add the new protocol fields, ordinals, particles, sounds, and memory categories.
- Align the new codecs with existing gophertunnel IO and Marshal patterns, removing transitional compatibility state and duplicate recipe/item codecs.

## Protocol references

- [CloudburstMC exact v2168 codec](https://github.com/CloudburstMC/Protocol/blob/19e31c33dc424b2b7e0098e0cbd9aae208705c2b/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/codec/v2168/Bedrock_v2168.java)
- [CloudburstMC v2168 gatherings codec](https://github.com/CloudburstMC/Protocol/blob/19e31c33dc424b2b7e0098e0cbd9aae208705c2b/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/codec/v2168/BedrockCodecHelper_v2168.java#L1166-L1191)
- [Mojang r/26_u4 changelog](https://github.com/Mojang/bedrock-protocol-docs/blob/7330880ab78ef001cad0b9cdfedb3aa3eaa6d4af/changelog_2168_07_07_26.md)
- [Mojang automated 1.26.40 gatherings schema](https://github.com/Mojang/bedrock-protocol-docs/blob/0e00fe80f4f3c71572ff6429de40146d1f4412fc/json/gatheringsConfig.json)
- [Mojang automated 1.26.40 sound-update schema](https://github.com/Mojang/bedrock-protocol-docs/blob/0e00fe80f4f3c71572ff6429de40146d1f4412fc/json/ClientboundUpdateSoundDataPacketPayload.json)
- [Mojang automated 1.26.40 Fade payload](https://github.com/Mojang/bedrock-protocol-docs/blob/0e00fe80f4f3c71572ff6429de40146d1f4412fc/json/Fade.json)
- [TwistedAsylumMC/bpd-fixer](https://github.com/TwistedAsylumMC/bpd-fixer)

The merged Mojang r/26_u4 metadata identifies protocol 2169 / Minecraft 1.26.50, while the pinned `automated/1.26.40` branch identifies protocol 2168 and is used for exact-version field order and union layout. In particular, every `ClientboundUpdateSoundData` slot is an independent seven-way union and `Fade` encodes `Duration` before `Target Volume`; Cloudburst's v2168 serializer hardcodes selector 0 and reverses those fields, so it is not followed for that packet.

Mojang's `gatheringsConfig` schema marks all eight fields required, and bpd-fixer currently leaves that metadata unchanged. Cloudburst's exact v2168 read/write codec instead shows optional markers for `worldId`, `worldName`, `targetId`, `scenarioId`, and `serverId`; this PR follows that wire codec while keeping `experienceId`, `experienceName`, and `creatorId` required.

## Verification

- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`
- Independent exact-version review against Mojang `automated/1.26.40`, Cloudburst v2168, and the corrected bpd-fixer schema
- GitHub Actions: passing

No new test files are included in this PR.


HashimTheArab:
@codex review

minecraft/protocol/packet/move_actor_delta.go
```diff
@@ -1,36 +1,30 @@
...
-	// Flags is a list of flags that specify what data is in the packet.
-	Flags uint16
 	// EntityRuntimeID is the runtime ID of the entity that is being moved. The packet works provided a
 	// non-player entity with this runtime ID is present.
 	EntityRuntimeID uint64
 	// Position is the new position that the entity was moved to.
-	Position mgl32.Vec3
+	PositionX protocol.Optional[float32]
+	PositionY protocol.Optional[float32]
+	PositionZ protocol.Optional[float32]
```
TrippleAWap, line 17: Even if Minecraft's internal protocol structures each position as inline floats show we really be doing the same? It seems kind of annoying and IMO it'd be better to just keep it as a Vec3 wrapping all of them together, just a thought.

HashimTheArab: It's annoying but the fields are independently optional. A single mgl32.Vec3 cannot represent “only Y was sent” without also restoring a separate flags/presence field.

TrippleAWap: I've thought about other solutions but they aren't very elegant and over complicate things for the user, although it's annoying I think you might be right; this seems to be best fitting.

minecraft/protocol/packet/set_scoreboard_identity.go
```diff
@@ -30,10 +30,23 @@ func (*SetScoreboardIdentity) ID() uint32 {
...
-		protocol.Slice(io, &pk.Entries)
-	case ScoreboardIdentityActionClear:
-		protocol.FuncIOSlice(io, &pk.Entries, protocol.ScoreboardIdentityClearEntry)
+	if pk.ActionType != ScoreboardIdentityActionRegister && pk.ActionType != ScoreboardIdentityActionClear {
+		io.UnknownEnumOption(pk.ActionType, "scoreboard identity action type")
+	}
+	protocol.FuncIOSlice(io, &pk.Entries, pk.marshalEntry)
+}
+
+func (pk *SetScoreboardIdentity) marshalEntry(io protocol.IO, entry *protocol.ScoreboardIdentityEntry) {
```
TwistedAsylumMC: Just thinking, should this be exported just so other people can call it?

minecraft/protocol/packet/set_scoreboard_identity.go
```diff
@@ -30,10 +30,23 @@ func (*SetScoreboardIdentity) ID() uint32 {
...
+	hasEntityUniqueID := expectsEntityUniqueID
+	io.Bool(&hasEntityUniqueID)
+	if hasEntityUniqueID != expectsEntityUniqueID {
+		io.InvalidValue(hasEntityUniqueID, "scoreboard identity entity unique ID presence", "does not match action type")
+	}
+	if hasEntityUniqueID {
+		io.Varint64(&entry.EntityUniqueID)
+	} else {
+		entry.EntityUniqueID = 0
 	}
```
TwistedAsylumMC: Why not just make it optional?

minecraft/protocol/packet/move_player.go
```diff
@@ -69,9 +66,15 @@ func (pk *MovePlayer) Marshal(io protocol.IO) {
...
-		io.Int32(&pk.TeleportCause)
-		io.Int32(&pk.TeleportSourceEntityType)
+	if _, writing := io.(*protocol.Writer); writing {
+		if pk.Mode == MoveModeTeleport {
+			if _, ok := pk.TeleportData.Value(); !ok {
+				pk.TeleportData = protocol.Option(protocol.TeleportData{})
+			}
+		} else {
+			pk.TeleportData = protocol.Optional[protocol.TeleportData]{}
+		}
```
TwistedAsylumMC: What is this?

minecraft/protocol/packet/resource_pack_client_response.go
```diff
@@ -28,6 +28,34 @@ func (*ResourcePackClientResponse) ID() uint32 {
...
-	protocol.FuncSliceUint16Length(io, &pk.PacksToDownload, io.String)
+	response := uint32(pk.Response)
+	io.Varuint32(&response)
+	if response > uint32(PackResponseCompleted) {
+		io.UnknownEnumOption(response, "resource pack response")
+	}
+	pk.Response = byte(response)
+
+	// A varuint32 length-prefixed string follows the response. Its contents are ignored by the client,
+	// so we write the readable name of the response for parity and discard whatever is read.
```
TwistedAsylumMC: Not needed

minecraft/protocol/packet/player_auth_input.go
```diff
@@ -5,7 +5,7 @@ import (
 	"github.com/sandertv/gophertunnel/minecraft/protocol"
 )
 
-const PlayerAuthInputBitsetSize = 65
+const PlayerAuthInputBitsetSize = 66
```
TwistedAsylumMC: Personally I don't think we should maintain this if it's no longer a bitset

minecraft/protocol/input_flags.go
```diff
@@ -0,0 +1,95 @@
...
+// An absent InputFlags is not the same as an empty one: it is not sent at all. The zero value is absent and
+// has no size, so use NewInputFlags before setting a flag on it. Decoded values are always sized.
+type InputFlags struct {
+	set  bool
+	size int
+	ids  []int32
+}
+
+// NewInputFlags creates an InputFlags holding flags in the range [0, size), with none of them set.
+func NewInputFlags(size int) InputFlags {
```
TwistedAsylumMC, line 17: Maybe add a second constructor for directly passing in the []int32 ids? I could see this being annoying otherwise having to call NewInputFlags() then Set() from a loop

minecraft/protocol/item_stack.go
```diff
@@ -296,6 +318,8 @@ const (
 	StackRequestActionDestroy
 	StackRequestActionConsume
 	StackRequestActionCreate
+	// StackRequestActionPlaceInContainer and StackRequestActionTakeOutContainer are never sent, but still
+	// take up their IDs in the action type the client repeats inside each action.
```
TwistedAsylumMC: Not needed

minecraft/protocol/map.go
```diff
@@ -52,12 +52,17 @@ type MapTrackedObject struct {
...
 		r.Varint64(&x.EntityUniqueID)
-	case MapObjectTypeBlock:
+	}
+	blockPresent := x.Type == MapObjectTypeBlock
+	r.Bool(&blockPresent)
+	if blockPresent {
 		r.BlockPos(&x.BlockPosition)
-	default:
+	}
+	if !entityPresent && !blockPresent {
```
TwistedAsylumMC: I'd rather there be optional fields on the struct itself rather than doing this. Sure only one should be set, but it's not really an issue if both are. It's down to whatever handles it to just ensure the type matches what data is present imo

minecraft/protocol/player.go
```diff
@@ -84,7 +94,28 @@ type PlayerListEntry struct {
...
+	r.Uint8(&legacyAction)
+	x.ActionType = PlayerListActionRemove
+	if variant == 1 {
+		x.ActionType = PlayerListActionAdd
+	} else if variant != 0 {
+		r.UnknownEnumOption(variant, "player list entry variant")
+	}
+	if legacyAction != x.ActionType {
+		r.InvalidValue(legacyAction, "player list action type", "does not match entry variant")
+	}
```
TwistedAsylumMC: Any way this could be cleaned up with a helper method somewhere?

minecraft/protocol/reader.go
```diff
@@ -268,14 +268,26 @@ func (r *Reader) UUID(x *uuid.UUID) {
 // PlayerInventoryAction reads a PlayerInventoryAction.
 func (r *Reader) PlayerInventoryAction(x *UseItemTransactionData) {
 	r.Varint32(&x.LegacyRequestID)
-	if x.LegacyRequestID < -1 && (x.LegacyRequestID&1) == 0 {
-		Slice(r, &x.LegacySetItemSlots)
+	var legacySlotsPresent bool
+	r.Bool(&legacySlotsPresent)
+	expectsLegacySlots := x.LegacyRequestID < -1 && (x.LegacyRequestID&1) == 0
```
TwistedAsylumMC: Not necessary imo

minecraft/protocol/reader.go
```diff
@@ -268,14 +268,26 @@ func (r *Reader) UUID(x *uuid.UUID) {
...
+		r.InvalidValue(legacySlotsPresent, "legacy set item slots presence", "does not match legacy request ID")
 	}
-	FuncSlice(r, &x.Actions, r.inventoryActionOld)
-	r.Varuint32(&x.ActionType)
-	r.Varuint32(&x.TriggerType)
+	if legacySlotsPresent {
+		Slice(r, &x.LegacySetItemSlots)
+	} else {
+		x.LegacySetItemSlots = nil
+	}
```
TwistedAsylumMC: just wrap the field in Optional?

minecraft/protocol/shape.go
```diff
@@ -112,6 +112,8 @@ type TextShape struct {
 	UseRotation bool
 	// BackgroundColour is the RGBA colour to use for the text background. This is a translucent black colour by default.
 	BackgroundColour Optional[color.RGBA]
+	// LineGapHeight is the gap between lines used when rendering multiline text.
+	LineGapHeight float32
```
TwistedAsylumMC: This change did not make it to U4

minecraft/protocol/skin.go
```diff
@@ -203,23 +242,37 @@ func (x *PersonaPiece) Marshal(r IO) {
...
 // Marshal encodes/decodes a PersonaPieceTintColour.
 func (x *PersonaPieceTintColour) Marshal(r IO) {
-	r.String(&x.PieceType)
-	FuncSliceUint32Length(r, &x.Colours, r.String)
+	wireType := personaPieceTintWireType(x.PieceType)
+	r.String(&wireType)
+	x.PieceType = personaPieceTintLoginType(wireType)
+	for i := range x.Colours {
+		r.BEARGB(&x.Colours[i])
+	}
```
TwistedAsylumMC, line 258: Should probably verify the size here so we dont end up writing less

minecraft/protocol/sub_chunk.go
```diff
@@ -45,36 +39,39 @@ type SubChunkEntry struct {
...
 func (x *SubChunkEntry) Marshal(r IO) {
 	Single(r, &x.Offset)
 	r.Uint8(&x.Result)
-	if x.Result != SubChunkResultSuccessAllAir {
+	hasRawPayload := x.RawPayload != nil
+	r.Bool(&hasRawPayload)
+	if hasRawPayload {
 		r.ByteSlice(&x.RawPayload)
+	} else {
+		x.RawPayload = nil
```
TwistedAsylumMC: Would it be easier to make these fields optional

TwistedAsylumMC (changes requested):

minecraft/protocol/packet/move_player.go
HashimTheArab: Simplified this by deriving the presence flag from Mode and validating it on decode. TeleportData is now the plain protocol payload.

minecraft/protocol/packet/resource_pack_client_response.go
HashimTheArab: Removed the comment.

minecraft/protocol/packet/player_auth_input.go
HashimTheArab: Renamed this to PlayerAuthInputFlagCount to reflect the list encoding.

minecraft/protocol/input_flags.go
HashimTheArab: Added NewInputFlagsFromIDs(size, ids) for bulk construction.

minecraft/protocol/item_stack.go
HashimTheArab: Removed the comment.

minecraft/protocol/player.go
HashimTheArab: Extracted the duplicated selector and legacy action handling into playerListAction.

minecraft/protocol/shape.go
HashimTheArab: Removed LineGapHeight and its encoding.

minecraft/protocol/sub_chunk.go
HashimTheArab: Converted the raw payload, height maps, render height maps, and blob hash to protocol.Optional fields.

minecraft/protocol/map.go
HashimTheArab: Changed both payloads on MapTrackedObject to Optional fields. Their presence is now represented independently from Type.

minecraft/protocol/packet/set_scoreboard_identity.go
HashimTheArab: Changed EntityUniqueID to protocol.Optional[int64] and marshal it with OptionalFunc.

minecraft/protocol/skin.go
HashimTheArab: Colours is now a [4]color.RGBA, so the type statically guarantees that exactly four colours are written.

minecraft/protocol/reader.go
HashimTheArab: Removed the LegacyRequestID-derived presence check; the wire marker is decoded directly into the optional field.

minecraft/protocol/reader.go
HashimTheArab: Wrapped both LegacySetItemSlots and Actions in protocol.Optional and use the existing optional helpers directly.

minecraft/protocol/packet/set_scoreboard_identity.go
HashimTheArab: Moved the entry codec onto protocol.ScoreboardIdentityEntry as its exported Marshal method, so the packet no longer needs a private packet-specific helper.

PR #466, opened 2026-07-12
Fixed float command argument type being displayed as "unknown"

JavierLeon9966:
Before:
Screenshot From 2026-07-12 04-38-41
After:
Screenshot From 2026-07-12 04-46-46

PR #487, opened 2026-08-05
protocol: correct command argument type constants

HashimTheArab:
The 1.26.33 client uses packet argument type ID `3` for float and ID `4` for value. ID `2` is unused.

This updates the constants accordingly:

- Keeps `CommandArgTypeInt` at `1`
- Reserves ID `2`
- Moves `CommandArgTypeFloat` to `3`
- Keeps `CommandArgTypeValue` at `4`
- Removes `CommandArgTypeRValue`, which is not a packet-facing argument type

This fixes float arguments being displayed as `unknown` without removing the valid `CommandArgTypeValue` constant.

### DIFF
--- a/minecraft/protocol/command.go
+++ b/minecraft/protocol/command.go
@@ -76,9 +76,9 @@ const (
 
 const (
 	CommandArgTypeInt = iota + 1
+	_
 	CommandArgTypeFloat
 	CommandArgTypeValue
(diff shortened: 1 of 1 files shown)


HashimTheArab:
tested on bds

HashimTheArab:
techinically #466 is correct

PR #491, opened 2026-08-06
Fix player list action constants for Bedrock 1.26.40

ItsMaipianX:
The PlayerList action constants were ordered incorrectly.

Bedrock protocol defines:
- PlayerListActionRemove = 0
- PlayerListActionAdd = 1

The previous implementation had PlayerListActionAdd = 0 and PlayerListActionRemove = 1 due to iota ordering, causing player list packets to be interpreted incorrectly by newer clients such as 1.26.40.

This changes the constants to match the protocol values.

### DIFF
--- a/minecraft/protocol/player.go
+++ b/minecraft/protocol/player.go
@@ -7,8 +7,8 @@ import (
 )
 
 const (
-	PlayerListActionAdd = iota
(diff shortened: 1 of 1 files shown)


THEBOSS9345:
Can you also next time show us the docs where it says that

PR #500, opened 2026-08-10
login/data.go: Fix validation for NetherNet ServerAddress field in ClientData.

Dasciam:


TwistedAsylumMC:
Can you show what the client is actually sending to make this necessary?

HashimTheArab:
i do this for lunar slightly different https://github.com/HashimTheArab/gophertunnel/blob/2c222c5e601c57826c8ab609d067f1286c997fb3/minecraft/protocol/login/data.go#L553-L566

lactyy2:
I think that is exactly the same as the current code. This PR adds support for `ServerAddress` with brackets (`[http://<host>:<port>]:<port>`), which I couldn't reproduce with any server address I tried.

Dasciam:
> I think that is exactly the same as the current code. This PR adds support for `ServerAddress` with brackets (`[http://<host>:<port>]:<port>`), which I couldn't reproduce with any server address I tried.

I've tried to just log in to a local server.

TwistedAsylumMC:
Can you show some data?

Dasciam:
> Can you show some data?

Caused by: https://github.com/Sandertv/gophertunnel/blob/4794743244247a9395360bced1f294d4882c4df0/minecraft/protocol/login/request.go#L172

Dasciam:
So, before ClientData.Validate call, the ServerAddress field is getting modified to the variant with brackets included

Got it

Dasciam:
Stale

PR #502, opened 2026-08-12
protocol: reject varints with set bits beyond the target width

HashimTheArab:
The varint readers terminate after 5/10 bytes, but never check the final byte's excess bits: a varuint32 whose fifth byte has any of the top four bits set (or a varuint64 whose tenth byte has any bit but the lowest set) decodes to a silently truncated value instead of failing. Vanilla never produces such encodings, so rejecting them only affects forged data. This adds the final-byte check to `protocol.Varuint32/Varuint64` and `Reader.Varuint32/Varuint64`.

Two small companions:

- `Reader.Varint32/Varint64` now decode through `Reader.Varuint32/Varuint64` and apply the zigzag themselves, matching how the package-level `Varint32/Varint64` already delegate — the overflow check lives in one place per width.
- `Reader.ByteSlice` caps length prefixes at 16 MiB via the existing `SliceLength` helper (which also keeps the remaining-payload check), so a forged prefix can no longer request an up-to-2 GB allocation before the payload length is ever compared.

### DIFF
--- a/minecraft/protocol/reader.go
+++ b/minecraft/protocol/reader.go
@@ -65,6 +65,10 @@ func (r *Reader) Bool(x *bool) {
 // errStringTooLong is an error set if a string decoded using the String method has a length that is too long.
 var errStringTooLong = errors.New("string length overflows a 32-bit integer")
 
+// maxByteSliceLength is the maximum length of a length-prefixed byte slice, so that a forged length prefix
+// cannot force an arbitrarily large allocation.
+const maxByteSliceLength = 16 * 1024 * 1024
+
 // StringUTF ...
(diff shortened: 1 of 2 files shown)


HashimTheArab:
> The varint changes are a pure correctness/security fix with extremely low regression risk. They reject only encodings that cannot fit the target integer width. Canonical vanilla encodings and non-canonical but still valid in-range encodings continue to work. Delegating signed decoding through unsigned decoding preserves zigzag behavior.

HashimTheArab:
cloudburst does the same as gophertunnels current behavior so ill reopen this pr if it becomes actually necessary

PR #506, opened 2026-08-16
protocol/player.go: Reorder player list action constants

ismaileke:


xRookieFight (approved):

TwistedAsylumMC:
This actually breaks stuff, we have the enum flipped in the encoding

PR #507, opened 2026-08-18
minecraft/nethernet.go: Use the same domain as the vanilla client for identity

lactyy2:
to match the vanilla client behavior, just in case


lactyy2:
...wrong branch

PR #523, opened 2026-09-18
minecraft/nethernet.go: Implement PingContext

lactyy2:
This PR implements the missing PingContext() method in `minecraft.NetherNet`. It also changes the code to use HTTP signaling by default when `NetherNet.Signaling` is nil.

`minecraft.NetherNet` is now registered as the `nethernet` network, so it can be used with `Listen` and `Dial` methods, for example:

```go
l, err := minecraft.Listen("nethernet", ":19132")
```


minecraft/nethernet.go
```diff
@@ -62,18 +66,84 @@ func (n NetherNet) DialContextIdentity(ctx context.Context, address string, toke
...
+	l, err := n.ListenConfig.Listen(n.Signaling)
+	if err != nil {
+		if closeFunc != nil {
+			if err2 := closeFunc(); err2 != nil {
+				err = errors.Join(err, fmt.Errorf("close HTTP server: %w", err2))
+			}
+		}
+		return nil, err
+	}
+	return netherNetListener{NetworkListener: l, close: closeFunc}, nil
```
HashimTheArab, line 118: Listening on `127.0.0.1:0` allocates an HTTP port, but `Addr()` exposes only the NetherNet network ID, so callers cannot discover that port. Could the wrapper preserve and expose the bound HTTP address when using default signaling?

